PR #2, opened 2013-09-02
Get build, execution and test for all platforms.

gmarkall:
- Modifications to the CMake UseJava module and addition of a
  NativeLibraryBuilder module so that add_jar can build JARs that
  contain native code that is part of the build tree. Native
  libraries to be included in the JAR are specified by calling:
  
    jar_native_library(<lib_target_name>)

in the CMakeLists.txt files.
- Fixes to the invocation of the tests on Windows for setting the
  classpath correctly.
- Inclusion of libogrpath from the old MathsWrappers library.
- Addition of a config file, NativeLibraries.properties, that
  specifies which libraries are to be extracted and loaded on each
  platform.
- A new native library loader that extracts and loads libraries
  specified in the configuration file.
- Exceptions to report error conditions in the native library
  loader.
- Materialisers and DOGMA2 call the native library initialisation
  code in order to ensure that native code is available at the
  time at which it is needed.
- Remove violation of the One-Definition Rule that was caused by
  an inclusion of bindings.hh in exprtypes.cc. The long-term fix
  for this problem will probably involve moving some definitions
  out of bindings.hh

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/6


gmarkall:
I've removed checkstyle and added a check for 0 bytes read.

New buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/7

PR #3, opened 2013-09-02
[MAT-159] Added OGResult, variadic return expr, updated C++ code to accomodate

stuartarchibald:
This is a rough implementation of [MAT-159]. It also renders [MAT-157] moot.
- Add SELECTRESULT AST node to allow selection of a result from a method that returns multiple results. This is not a list impl but a custom class as discussed in the MAT-159 ticket commentary.
- Add OGResult.java which allows users to access data from classes which have multiple returned items. This code simply adds in suitable SELECTRESULT nodes into the AST that hold pointers to the Node from which the selection shall take place.
- Remove SelectEnum enumerations in favour of holding an int reference to the selected index.
- Add SVD reference impl.
- A few debug bits added.
- Corrected erroneous error message in MathsException.java

BB pass following merge of master which contains #2 http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/8


include/exprtypeenum.h
```diff
@@ -26,34 +26,17 @@
...
 
-  PLUS_ENUM    = 0X0100L ,
-  MINUS_ENUM   = 0X0200L ,
-  TIMES_ENUM   = 0X0200L ,
-  LDIVIDE_ENUM = 0X0300L ,
-  MTIMES_ENUM  = 0X0500L ,
-  COPY_ENUM    = 0X0600L ,
+  PLUS_ENUM            = 0X0100L ,
+  MINUS_ENUM           = 0X0200L ,
+  TIMES_ENUM           = 0X0200L ,
```
gmarkall: why do MINUS and TIMES have the same values, and why os 0X0400L skipped?

stuartarchibald: because I can't count :) patched d7a4006.

gmarkall:
A couple of questions:
1. Why do we need to have OGResult being distinct from OGExprVariadicReturn? If OGExprVariadicResult had the get method, it  looks like OGResult woudn't be needed, and it wouldn't be necessary to wrap an SVD object in an OGResult in DOGMA2.svd.
2. Do we want to allow things like:

``` language=java
a = svd(m1);
b = svd(m2);
c = plus(a, b);
```

in which case each element of the SVD results is added to the corresponding element from the other SVD?
1. Master has been merged in to this branch - merging master into feature branches has been a source of pain when rewriting in the past (and I think the rule for the platform is to only rebase feature branches too, but I think it's often broken) - could this branch be rebased on master instead of master being merged into it?
2. Buildbot seems to pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/9

stuartarchibald:
1. Agreed. Think this highlights the roughness of the impl. Result of starting out with List<T> and swapping. I'll sort this tomorrow as it's poor, will see if anything comes up that stops it just working as suggested.
2. Yes, that operation is conceivable though perhaps not massively useful. Is this a general query or is there something more deep I'm missing? As java can't differentiate on return type and svd() needs the capacity to return multiple items, 'a' and 'b' will need to be more like:

``` language=java
a = svd(m1).get(0);
b = svd(m2).get(0);
c = plus(a, b);
```

is this going to give us problems do you think?
1. Yeah, bad practice, sorry. I'll revoke and rebase.
2. Yep, I updated the wrong thing on github comment so I think we've both run that SHA and it definitely passes.

gmarkall:
[ This is what I currently started to think as a response, but I think it needs more discussion, so I'm posting it as-is so I can refer to it ]

(I can count btw, but Github has it's own idea of how to count and re-numbered my list that was 1-4).
1. I'm beginning to have second thoughts about whether OGResult should be removed entirely - see point 2.
2. It was more of a general query, partly because I haven't thought ahead far enough to foresee if allowing that kind of thing might cause us trouble. 

However, with the example that we are discussing, if I add types to my first example:

``` language=java
OGExpr svd(m1); // (where DOGMA2.svd returns svd, which is OGExprVariadicReturn, which is OGExpr)
OGExpr svd(m2); // as above
OGExpr c = plus(a, b); // Since a and b are instances of OGExpr, there's no need to differentiate on type
```

I think this doesn't need any differentiation on the return type - perhaps the OGResult class would have been a problem, since it's outside the OGExpr hierarchy?

Now, I think that in order to actually use the variable `c`, you'd have to do something like:

``` language=java
c.get(0); // is U1+U2
c.get(1); // is S1+S2
c.get(2); // is V1+V2
```

Thinking further about it, this interferes with the current design, since plus is not an OGExprVariadicResult. Off the top of my head there are a couple of possibilities: 
1. Doing computation directly with OGExprVariadicResult objects could be prohibited
2. Everything could be variadic, and most operations return a single result. In this case, I think the OGResult class starts to make sense (except that I'd call it something like OGExprList or OGListExpr, and it would be a subclass of

stuartarchibald:
Resolved via verbal discussion. OGResult stays as it forces correctness at compile time. 
List comprehension style operations are not being added for now for fear of incorrect use and possible cries of "that's not what's expected".
Passes: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/10

PR #5, opened 2013-09-03
Compile with warnings/errors

gmarkall:
Add -Wall -Wextra -Werror -pedantic flags to build, and set flags for each build type appropriately.

Errors detected with these flags included several missing returns, unsafe comparisons, etc. 

The set of changes included fixes all the errors that were highlighted by these flags.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/26

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,15 +37,15 @@ set(longdog_VERSION
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules ${CMAKE_MODULE_PATH})
 
 # tweak flags
-set(CMAKE_C_FLAGS "-ffast-math -O2 -march=core2 -funroll-loops ${CMAKE_C_FLAGS}")
-set(CMAKE_C_FLAGS_DEBUG "-ffast-math -O2 -march=core2 -funroll-loops ${CMAKE_C_FLAGS_DEBUG}")
-set(CMAKE_C_FLAGS_RELEASE "-ffast-math -O2 -march=core2 -funroll-loops ${CMAKE_C_FLAGS_RELEASE}")
-set(CMAKE_C_FLAGS_RELWITHDEBINFO "-ffast-math -O2 -march=core2 -funroll-loops ${CMAKE_C_FLAGS_RELWITHDEBINFO}")
+set(CMAKE_C_FLAGS "-std=c99 -Wall -Werror -Wextra -pedantic ${CMAKE_C_FLAGS}")
+set(CMAKE_C_FLAGS_DEBUG "-DDEBUG ${CMAKE_C_FLAGS_DEBUG}")
+set(CMAKE_C_FLAGS_RELEASE "-ffast-math -O2 -march=core2 -funroll-loops -DNDEBUG")
+set(CMAKE_C_FLAGS_RELWITHDEBINFO "-ffast-math -march=core2 -funroll-loops ${CMAKE_C_FLAGS_RELWITHDEBINFO}")
 
-set(CMAKE_CXX_FLAGS "-std=c++11 -ffast-math -O2 -march=core2 -funroll-loops ${CMAKE_CXX_FLAGS}")
-set(CMAKE_CXX_FLAGS_DEBUG "-std=c++11 -ffast-math -O2 -march=core2 -funroll-loops ${CMAKE_CXX_FLAGS_DEBUG}")
-set(CMAKE_CXX_FLAGS_RELEASE "-std=c++11 -ffast-math -O2 -march=core2 -funroll-loops ${CMAKE_CXX_FLAGS_RELEASE}")
-set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-std=c++11 -ffast-math -O2 -march=core2 -funroll-loops ${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
+set(CMAKE_CXX_FLAGS "-std=c++11 -Wall -Werror -Wextra -pedantic ${CMAKE_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS_DEBUG "-DDEBUG ${CMAKE_CXX_FLAGS_DEBUG}")
+set(CMAKE_CXX_FLAGS_RELEASE "-ffast-math -O2 -march=core2 -funroll-loops -DNDEBUG")
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-ffast-math -march=core2 -funroll-loops ${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
 
 
 # Add Java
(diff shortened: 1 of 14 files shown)


gmarkall:
Forgot to rebase before I made the pull request. Here is the buildbot pass for the rbased branch: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/27

src/jshim/jshim.cc
```diff
@@ -691,7 +699,7 @@ void * instantiateJClassAsCXXClass(jobject obj)
    * Signature: (Lcom/opengamma/longdog/datacontainers/OGNumeric;)Lcom/opengamma/longdog/datacontainers/OGNumeric;
    */
   JNIEXPORT jobject JNICALL Java_com_opengamma_longdog_materialisers_Materialisers_materialise
-  (JNIEnv * env, jclass clazz, jobject obj)
+  (JNIEnv __attribute__ ((unused)) *env, jclass __attribute__ ((unused)) clazz, jobject obj)
```
stuartarchibald: Temporarily unused yes, can't imagine it'll remain so. Presume this was to silence warnings. Perhaps a  `#define SUPPRESS_UNUSED __attribute__ ((unused))` ?

src/jshim/jshim.cc
```diff
@@ -703,7 +711,8 @@ void * instantiateJClassAsCXXClass(jobject obj)
 
     printf("Calling entrypt function\n");
     // pass to entrypt
-    struct c_OGNumeric * answer = entrypt((struct c_OGNumeric *) chain);
+    //struct c_OGNumeric * answer = (should answer be used later?)
```
stuartarchibald, line 715: Yes it probably should. This relates to MAT-175.

src/ogrpath/rpath_hack.c
```diff
@@ -19,13 +19,15 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *jvm, void *reserved)
...
+int varInThisLib = 0;
+
 #ifdef __cplusplus
 extern "C"
 #endif
 // gets the handle to *the current* module, use address of this lib and don't increment ref count
 HMODULE getThisLibHandle() {
     HMODULE lib = NULL;
-    GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS|GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, (LPCTSTR)getThisLibHandle, &lib);
+    GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS|GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, (LPCTSTR)&varInThisLib, &lib);
```
stuartarchibald, line 31: Purely out of interest, what warning did this give?

CMakeLists.txt
```diff
@@ -86,15 +86,21 @@ message("Build type: " ${CMAKE_BUILD_TYPE})
...
+  C Flags (release)..........: " ${CMAKE_C_FLAGS_RELEASE} "
+  C Flags (debug)............: " ${CMAKE_C_FLAGS_DEBUG} "
+  C Flags (relwithdebinfo)...: " ${CMAKE_C_FLAGS_RELWITHDEBINFO} "
+  C++ Compiler...............: " ${CMAKE_CXX_COMPILER} "
+  C++ Flags..................: " ${CMAKE_CXX_FLAGS} "
+  C++ Flags (release)........: " ${CMAKE_CXX_FLAGS_RELEASE} "
+  C++ Flags (debug)..........: " ${CMAKE_CXX_FLAGS_DEBUG} "
+  C++ Flags (relwithdebinfo).: " ${CMAKE_CXX_FLAGS_RELWITHDEBINFO} "
+  Java Compiler..............: " ${Java_JAVAC_EXECUTABLE} "
+  Java Compiler flags........: " ${CMAKE_Java_COMPILE_FLAGS} "
```
stuartarchibald, line 101: should "Java" be CAPS for consistency?

include/numerictypes.hh
```diff
@@ -0,0 +1,21 @@
+/*
+ * numerictypes.hh
+ *
+ *  Created on: 3 Sep 2013
+ *      Author: graham
+ */
+
+#ifndef NUMERICTYPES_HH_
+#define NUMERICTYPES_HH_
```
stuartarchibald: - Are we going with NAME_HH_ or _NAME_HH ? we have a mix.
- Author and creation tag need to go, replace with copyright (MAT-179 is the ticket for that)

include/visitor.hh
```diff
@@ -0,0 +1,41 @@
+/*
+ * visitor.hh
+ *
+ *  Created on: 3 Sep 2013
+ *      Author: graham
+ */
+
```
stuartarchibald: - Author and creation tag need to go, replace with copyright (MAT-179 is the ticket for that)

src/jshim/jshim.cc
```diff
@@ -83,9 +83,9 @@ class JOGRealMatrix: public OGRealMatrix
       printf("\nJava bound OGRealMatrix\n");
       size_t ptr=0;
       printf("\n");
-      for(size_t i = 0 ; i < this->getRows(); i++)
```
stuartarchibald, line 86: Whilst with java we can get away with `int` for everything, for the pure C++ side do you think `size_t` or some indexing type we define might give us more scope esp wrt "large" arrays?

src/jshim/jvmcache.c
```diff
@@ -217,12 +224,19 @@ jint registerGlobalMethodReference(JNIEnv * env, jclass * globalRef, jmethodID *
...
 #ifdef _DEBUG
-    printf("Method found %s() 0x%llx\n",methodName,methodToSet);
+#ifdef __MINGW32__
+  unsigned int high, low;
+  low = (long long unsigned int)methodToSet & 0x00000000FFFFFFFFULL;
+  high = ((long long unsigned int)methodToSet & 0xFFFFFFFF00000000ULL) >> 32;
+  printf("Method found %s() 0x%x%x\n", methodName, high, low);
+#else
+    printf("Method found %s() 0x%llx\n",methodName,(long long unsigned int)methodToSet);
+#endif
```
stuartarchibald: This appears twice. Macroize?

src/librdag/visitor.cc
```diff
@@ -0,0 +1,16 @@
+/*
+ * visitor.cc
+ *
+ *  Created on: 3 Sep 2013
+ *      Author: graham
+ */
+
```
stuartarchibald: As previous comments on authorship etc.

src/jshim/jshim.cc
gmarkall: It was to silence the errors, yes. I can add the SUPPRESS_UNUSED macro, or **attribute** ((unused)) can just be shortened to **unused** - which would you find preferable?

gmarkall: (double underscore interepreted by Github as bold text)

src/jshim/jshim.cc
gmarkall, line 715: OK - I'll leave this as-is for now then.

src/ogrpath/rpath_hack.c
gmarkall, line 31: "ISO C forbids the conversion of a function pointer into an object instance pointer" or something similar.

src/jshim/jshim.cc
stuartarchibald: Think we'll need more suppressions? In which case lets add `SUPPRESS_FOO` else I'm happy with `__unused` if you are.

CMakeLists.txt
gmarkall, line 101: Unfortunately there are parts of the CMake code (its C++ source) that refer to Java in a case-sensitive way, so unfortunately we can't change that. Most things work if you use JAVA, but some oddity can occur.

include/numerictypes.hh
gmarkall: Whoops, that's because I used an Eclipse template to generate the file. I prefer _NAME_HH, since we mostly use the pattern of underscore first to keep things "private" in some sense. Will address your comment accordingly.

include/visitor.hh
gmarkall: Will do, as above

src/ogrpath/rpath_hack.c
stuartarchibald, line 31: Ah, potential size difference.

CMakeLists.txt
stuartarchibald, line 101: Fair enough. I guess leave unless consistency can easily be forced upstream.

src/jshim/jshim.cc
gmarkall, line 86: size_t is an unsigned type, but rows, cols and datalen are signed. I left them signed for the time being in case we used -1 as an initial value, but I'd definitely prefer using unsigned types for things like this would be better.

If you think we can avoid using -1 as the initial value for these variables, then I'd be keen to change all these induction variables so they have unsigned types. (If we're going to use size_t for the induction variables, then I'd make rows, cols and datalen size_t as well).

src/jshim/jvmcache.c
gmarkall: OK

src/librdag/visitor.cc
gmarkall: OK

src/jshim/jshim.cc
gmarkall: I don't expect us needed more suppressions - I think they're just necessary here because the interface is complete before the implementation is. I'll change to `__unused__` for now - I think in general if the error about unused parameters comes up, its generally a sign that the interface is wrong or the implementation is missing something, and it would be more preferable to get the interface right or fix the implementation, than just suppress the warning.

src/jshim/jshim.cc
stuartarchibald, line 86: Think that's just my bad habits, fortran style debug ;-)

I see no problem with the C++ side having size_t or some other more preferable unsigned integer type esp. for cases of induction. I think the only\* problem is going to happen when things are run through the jshim layer as there's potential for things like:
`jint foo = size_t bar;` (possibly interpreted as) `int foo = unsigned int bar;`
and the reverse.

*assuming we have wide enough unsigned types on all systems, which for 64bit JVMs I'd hope we're ok.

thoughts?

src/jshim/jshim.cc
stuartarchibald: I'm happy with that, thanks.

src/jshim/jshim.cc
gmarkall, line 86: I've just checked, and size_t is 64 bits on all platforms. I believe jint varies between long int and long long int on the platforms, but I think they all mean 64 bit as well. So I don't think there should be a problem with the sizes, but I can see how we could have overflow because of the signedness on the JVM side (we'd get errors if we just wrote the code in the naive way, but we could cast and then get the potential for overflow). 

The only solution to that that I can think of is to check that integers are positive when they come in from the JVM size, and to make sure they don't exceed INT64_MAX (or whatever it's called) when passing numbers back to the JVM. Maybe we could surround those checks with ifdef DEBUG macros, since it's not likely that we'll see anything that large any time soon?

gmarkall, line 86: > I believe jint varies between long int and long long int on the platforms, but I think they all mean 64 bit as well

When I said that I was confused between jint and jlong. Please ignore it.

gmarkall:
I think I've resolved all the issues you pointed out, with the exception of choosing the int type, which I think is a task for another PR, since we're not 100% sure what to do yet.

Are there any other issues you'd like me to resolve?

Buildbot pass for most recent revision: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/30

PR #7, opened 2013-09-05
Addition of testing and removal of Diamond

gmarkall:
This pull request consists of:
1. Setup of CMake for using the Google Test framework, and some rudimentary tests to ensure that it works.
2. Various refactoring/bugfix of previously-untested functionality/implementation of missing but useful functions in the expression hierarchy
3. Removal of the diamond inheritance in jshim. Each class has a factory function that constructs itself and then uses JOGExpr to construct its arguments, which it can then set on itself. This is because JOGExpr can't access setArgs() of OGExpr anymore. The factory functions can be refactored to reduce duplication, but I'm making this pull request sooner than that so that we are working on more similar interfaces before diverging too far.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,7 @@ set(CMAKE_JAVA_INCLUDE_PATH $ENV{CLASSPATH})
 # Place executables and libraries in bin on Windows so that its linker can find them.
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${longdog_BINARY_DIR}/bin)
 
+set(GTEST_ROOT $ENV{GTEST_ROOT})
 find_package(GTest REQUIRED)
 include_directories(${GTEST_INCLUDE_DIRS})
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,9 +65,13 @@ set(CMAKE_JAVA_TARGET_OUTPUT_DIR ${JARDIR})
 
 set(CMAKE_JAVA_INCLUDE_PATH $ENV{CLASSPATH})
 
-add_subdirectory(src)
+# Place executables and libraries in bin on Windows so that its linker can find them.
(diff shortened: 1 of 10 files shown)


gmarkall:
Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/47

CMakeLists.txt
```diff
@@ -65,9 +65,17 @@ set(CMAKE_JAVA_TARGET_OUTPUT_DIR ${JARDIR})
 
 set(CMAKE_JAVA_INCLUDE_PATH $ENV{CLASSPATH})
 
-add_subdirectory(src)
+# Place executables and libraries in bin on Windows so that its linker can find them.
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${longdog_BINARY_DIR}/bin)
+
+set(GTEST_ROOT $ENV{GTEST_ROOT})
+find_package(GTest REQUIRED)
+include_directories(${GTEST_INCLUDE_DIRS})
```
stuartarchibald, line 73: Do you think its worth having the ability to compile the libraries without the test suite harnesses being available?

stuartarchibald:
Am happy to merge this, refactor looks good.

CMakeLists.txt
gmarkall, line 73: The only "safe" way to use Google Test is to include it in your source repository and build it along with your code, so that it inherits all your compiler flags (and possibly anything else CMake is configured to do). I don't believe this is so much of a problem on Linux, but the proliferation of calling conventions on Mac and Windows can lead to trouble.

I had as yet avoided adding the Google Test code to the repository until it is clear that we want to go with it. However, if we do stick with it and include the source in a separate subfolder, then the presence of it on anyone's system stops being an issue.

I could commit this now if it'd be less hassle for you in the immediate-term - I'm also becoming convinced that it's good for our purposes since it seems to work on all platforms, be BSD licensed, and seems to make it simple to write tests and fixtures.

PR #8, opened 2013-09-06
Set up Eclipse templates and reformat code.

gmarkall:
I'm making this PR as a starting point for discussion. I set up a code template in Eclipse that was as similar to the general style we've been using as I could get it to be, then used that to format every source file.

What are the initial impressions of this formatting? Are there things that need changing?

I'm thinking that this might be suitable to merge after the visitor/toJDoubleArray stuff goes in so that the rebase isn't a terrible pain.


include/exprtypeenum.h
```diff
@@ -10,33 +10,33 @@
...
+   */
+  OGREALSCALAR_ENUM = 0x0001L,
+  OGCOMPLEXSCALAR_ENUM = 0X0002L,
+  OGREALMATRIX_ENUM = 0X0003L,
+  OGCOMPLEXMATRIX_ENUM = 0X0004L,
+  OGREALSPARSEMATRIX_ENUM = 0X0005L,
+  OGCOMPLEXSPARSEMATRIX_ENUM = 0X0006L,
+  OGREALDIAGONALMATRIX_ENUM = 0X0007L,
+  OGCOMPLEXDIAGONALMATRIX_ENUM = 0X0008L,
+  OGLOGICALMATRIX_ENUM = 0X0009L,
```
gmarkall, line 27: There seems to be no way around this stupid formatting. And no way to temporarily disable the formatter like in Java.

gmarkall:
Closing since neither of us is using Eclipse for C++.

PR #10, opened 2013-09-09
Mat 175 rebase

stuartarchibald:
Addresses MAT-175, MAT-186, MAT-187.
BB PASS: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/57
- [MAT-175] Work out to instantiate OGNumeric types back into the JVM
  - Abandoned the idea of instantiating terminal types into JVM, went for raw types instead. This reduces work.
  - This is WIP, toJBoolean still needs writing
  - this has lead to additional tickets and work related to the extensibility and implementation of the Visitor class [MAT-191]
  - Disabled a few java tests that essentially aren't tests but just scratch space/debug type tests that currently fail.
- [MAT-187] Move all exceptions to /exceptions
- [MAT-188] Add OGTerminal node
  - Impl adds pure virtual to<Real16,Complex16>ArrayOfArrays() to all terminals. This is not a template class as it is materialised to an `lvalue` the type of which cannot be determined at compile time.


include/expression.hh
```diff
@@ -25,6 +25,18 @@ using namespace std;
...
+ * General execption thing
+ */
+class librdagException: public exception
+{
+    public:
+    virtual const char* what() const throw()
+    {
+      return "Exception occurred.";
+    }
+};
```
gmarkall: Should exception live in their own header rather than being mixed in with expressions?

gmarkall: Additionally, I think this violates the ODR if the implementation is in the header, since it'll be implemented in every translation unit.

include/expression.hh
```diff
@@ -243,7 +307,22 @@ template <typename T> class OGMatrix: public OGArray<T>
...
+      for(int i=0; i < rows; i++)
+      {
+        tmp[i] = new T [cols];
+        for(int j = 0; j < cols; j++) {
+          tmp[i][j] = data[j*rows+i];
+        }
+      }
+      return tmp;
+    }
+private:
```
gmarkall: Stray private keyword?

stuartarchibald: fix: 79fd24e . BB Pass http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/61 (includes a407395)

include/expression.hh
stuartarchibald: yep. Many ODR violations. Best not create more. Fixed a407395. BB Pass http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/61

src/jshim/jshim.cc
```diff
@@ -707,47 +729,357 @@ void * instantiateJClassAsCXXClass(jobject obj)
...
+     return this->cols; 
+    }    
+  private:
+    complex16 ** _data = NULL;
+    int rows;
+    int cols;
+};
+
+
+  /**
```
gmarkall, line 897: What will deleteArrOfArr be used for?

stuartarchibald, line 897: I was using it in the destructors of the DispatchTo<T>16ArrayOfArrays, clearly not any more, would you prefer it removed? I expect at some point things like that will need to go into a Utils library.

gmarkall, line 897: I think it's fine left as it is - I was just curious.

PR #11, opened 2013-09-10
[MAT-196] Make arguments of expressions immutable.

gmarkall:
This is done using the PtrVector class, which holds an STL vector
but only permits its contents to be read, not modified, and does
not allow null pointers to be inserted.

Presently only wrapper methods that are used are implemented.
Others may need to be implemented to get to other functionality of
the underlying vector in the future.

The consquence of making its contents immutable (and returning
only const pointers to its elements) is that dynamic_cast is not
possible because it would cast away the const-ness. In order to
get around this, a dynamic_cast-like mechanism is implemented for
the OGNumeric class.

So far this is only implemented for the scalar types: OGNumeric
has members asOG*Scalar, which are virtual and return null. The
derived class then overrides the method that mentions its name
(e.g. OGRealScalar overrides asOGRealScalar) so that it returns
the this pointer.

Thus, the effect of calling OGNumeric::asOGRealScalar() is the
same as using dynamic_cast<OGRealScalar*>(ptr), except that it
uses const pointers, and will work across module boundaries,
unlike C++ RTTI.

### DIFF
--- a/src/jshim/jshim.cc
+++ b/src/jshim/jshim.cc
@@ -364,7 +364,7 @@ class JOGExpr
   private:
     jobject * _backingObject = NULL;
   protected:
-    std::vector<OGNumeric *> *generateArgs();
+    ArgContainer* generateArgs();
 };
 /**
  * COPY node spec derived from a java COPY node
@@ -564,7 +564,7 @@ JOGExpr::JOGExpr(jobject * obj)
   this->_backingObject = obj;
 }
 
-std::vector<OGNumeric *> *
+ArgContainer*
 JOGExpr::generateArgs()
 {
   jobject *obj = this->_backingObject;
@@ -601,7 +601,7 @@ JOGExpr::generateArgs()
 #ifdef _DEBUG
   printf("JOGExpr arg size is %d\n",len);
 #endif
-  std::vector<OGNumeric *> * local_args = new std::vector<librdag::OGNumeric *>;
+  ArgContainer* local_args = new ArgContainer();
   ExprFactory * factory = new ExprFactory();
   for(int i=0; i<len; i++)
   {
--- a/src/librdag/expression.cc
+++ b/src/librdag/expression.cc
@@ -28,6 +28,21 @@ void OGNumeric::debug_print()
   cout << "Abstract OGNumeric type" << endl;
 }
 
+const OGRealScalar* OGNumeric::asOGRealScalar() const
+{
+  return nullptr;
+}
+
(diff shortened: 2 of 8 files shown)


stuartarchibald:
BB pass? warningmacros.h is missing (from src) AFAICS?!

gmarkall:
Buildbot pass is: http://devmaths-lx-1:8011/builders/nyqwk2-testing-lnx-Debug/builds/68

warningmacros.h is in master already.

src/jshim/jshim.cc
```diff
@@ -601,7 +601,7 @@ class ExprFactory
 #ifdef _DEBUG
   printf("JOGExpr arg size is %d\n",len);
 #endif
-  std::vector<OGNumeric *> * local_args = new std::vector<librdag::OGNumeric *>;
+  ArgContainer* local_args = new ArgContainer();
```
stuartarchibald, line 604: delete local_args; at some point ??

src/librdag/expression.cc
```diff
@@ -39,31 +54,27 @@ void OGNumeric::debug_print()
 
 OGExpr::OGExpr(OGExpr& copy)
 {
-  this->_args = new std::vector<OGNumeric *>(*copy.getArgs());
+  this->_args = new ArgContainer(*copy.getArgs());
```
stuartarchibald, line 57: deletes ? I know they aren't there for vector<T * > but they should be.

stuartarchibald:
ah yes, warningmacros.h is there, must have just missed it.

src/librdag/expression.cc
gmarkall, line 57: When you delete an ArgContainer it deletes its elements as well, which is why the deletes were removed from OGExpr. OGExpr can no longer delete its own arguments, because it is not allowed to modify them through the ArgContainer.

src/jshim/jshim.cc
gmarkall, line 604: I think so - I'll generally sort this function out as part of MAT-198.

gmarkall, line 604: Unless i misunderstood you - are you suggesting there's a memory leak? local_args belongs to the OGExpr node that is passed to, which deletes it itself on destruction.

gmarkall:
Updated the tests now to test each member of PtrVector.

Suppressions have been removed as they're unnecessary - no memory leaks are detected in librdag with these tests.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/69

stuartarchibald:
Happy to merge. Does as expected.

PR #13, opened 2013-09-10
Mat 161 - add diagonal matrix support through JNI

stuartarchibald:
This PR adds support for diagonal matrix classes.
- Fixes up prototype diagonal matrix C++ code.
- Slightly alters internal storage and `<cinit>` behaviour in java as C/C++ doesn't know about the data lengths, the cost of this change in terms of algorithmic complexity is trivially small as the data sets are essentially vectors. The change essentially zero pads a short diagonal to min(rows,cols) so that loop bounds and algs are cleaner.
- Adds our first end to end test of a java `OGTerminal` `java -> C++ -> java` round trip. This is done as a materialise on `OG*Matrix` and `OG*DiagonalMatrix` classes.

bb pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/73](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/73)

### DIFF
--- a/src/jshim/jshim.cc
+++ b/src/jshim/jshim.cc
@@ -450,67 +450,89 @@ class ExprFactory
       {
       case OGREALMATRIX_ENUM:
       {
+#ifdef _DEBUG        
         printf("Binding a JOGRealMatrix\n");
+#endif        
         _expr = new JOGRealMatrix(&obj);
       }
       break;
       case OGCOMPLEXMATRIX_ENUM:
       {
+#ifdef _DEBUG        
         printf("Binding a JOGComplexMatrix\n");
+#endif                
         _expr = new JOGComplexMatrix(&obj);
       }
       break;
       case OGREALSPARSEMATRIX_ENUM:
       {
+#ifdef _DEBUG        
         printf("Binding a JOGRealSparseMatrix\n");
+#endif                
         _expr = new JOGRealSparseMatrix(&obj);
       }
       break;
       case OGCOMPLEXSPARSEMATRIX_ENUM:
       {
+#ifdef _DEBUG        
         printf("Binding a JOGComplexSparseMatrix\n");
+#endif                
         _expr = new JOGComplexSparseMatrix(&obj);
       }
       break;
       case OGREALDIAGONALMATRIX_ENUM:
       {
+#ifdef _DEBUG        
         printf("Binding a JOGRealDiagonalMatrix\n");
+#endif                
         _expr = new JOGRealDiagonalMatrix(&obj);
       }
       break;
       case OGCOMPLEXDIAGONALMATRIX_ENUM:
       {
+#ifdef _DEBUG        
         printf("Binding a JOGComplexDiagonalMatrix\n");
+#endif                
         _expr = new JOGComplexDiagonalMatrix(&obj);
       }
       break;      
       case COPY_ENUM:
       {
+#ifdef _DEBUG        
         printf("COPY function\n");
+#endif                
         _expr = new COPY(generateArgs(&obj));
       }
       break;
       case MINUS_ENUM:
       {
+#ifdef _DEBUG          
         printf("MINUS function\n");
+#endif                
         _expr = new MINUS(generateArgs(&obj));
       }
       break;
       case PLUS_ENUM:
       {
+#ifdef _DEBUG          
         printf("PLUS function\n");
+#endif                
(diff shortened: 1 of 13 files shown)


src/jshim/jshim.cc
```diff
@@ -426,6 +476,20 @@ class ExprFactory
         _expr->debug_print();
       }
       break;
+      case OGREALDIAGONALMATRIX_ENUM:
+      {
+        printf("Binding a JOGRealDiagonalMatrix\n");
+        _expr = new JOGRealDiagonalMatrix(&obj);
+        _expr->debug_print();
```
gmarkall: Are the debug_prints still required?

src/jshim/jshim.cc
```diff
@@ -426,6 +476,20 @@ class ExprFactory
...
+      }
+      break;
+      case OGCOMPLEXDIAGONALMATRIX_ENUM:
+      {
+        printf("Binding a JOGComplexDiagonalMatrix\n");
+        _expr = new JOGComplexDiagonalMatrix(&obj);
+        _expr->debug_print();
```
gmarkall: Another debug_print

src/jshim/jshim.cc
```diff
@@ -646,6 +710,7 @@ class DispatchToComplex16ArrayOfArrays: public librdag::Visitor
     };
     void visit(librdag::OGArray<complex16> *thing)
     {
+      printf("Visitor: librdag::OGArray<complex16> branch\n");
```
gmarkall, line 729: Debugging printf

src/jshim/jshim.cc
```diff
@@ -656,6 +721,7 @@ class DispatchToComplex16ArrayOfArrays: public librdag::Visitor
     }
     void visit(librdag::OGMatrix<complex16> *thing)
     {
+      printf("Visitor: librdag::OGMatrix<complex16> branch\n");
```
gmarkall, line 740: Debugging printf

src/jshim/jshim.cc
```diff
@@ -426,6 +476,20 @@ class ExprFactory
...
+      case OGREALDIAGONALMATRIX_ENUM:
+      {
+        printf("Binding a JOGRealDiagonalMatrix\n");
+        _expr = new JOGRealDiagonalMatrix(&obj);
+        _expr->debug_print();
+      }
+      break;
+      case OGCOMPLEXDIAGONALMATRIX_ENUM:
+      {
+        printf("Binding a JOGComplexDiagonalMatrix\n");
```
gmarkall, line 494: debug printf

src/jshim/jshim.cc
```diff
@@ -426,6 +476,20 @@ class ExprFactory
         _expr->debug_print();
       }
       break;
+      case OGREALDIAGONALMATRIX_ENUM:
+      {
+        printf("Binding a JOGRealDiagonalMatrix\n");
```
gmarkall, line 486: debug printf

src/jshim/jshim.cc
stuartarchibald: Nope.
Fix: b66d650
BB Pass. http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/77

src/jshim/jshim.cc
stuartarchibald: See b66d650

gmarkall:
This looks generally good - is it a good time to remove or #ifdef out those debugging prints? (I think i noted them all, with the intent of making your life easier if you do want to remove them).

Also, I notice that the OGDiagonalMatrix has a constructor that copies the data, and a public copy constructor that copies the data - I seem to recall that we concluded that copies shoudn't copy data, otherwise we'll make multiple copies when manipulating ASTs - do you agree? (if you do, I'm not suggesting changing it in this PR, but rather that we have a ticket for sorting that out to fix another time).

src/jshim/jshim.cc
stuartarchibald, line 486: These exist in all `case` s. Shall we remove totally or wrap them with #ifdef _DEBUG?

stuartarchibald:
Yes so as noted there are debug `printf`s in all `case` s on that switch. I'm happy to wrap with debug.

stuartarchibald:
489311a wraps debug printf statements.
BB Pass.  http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/78

PR #14, opened 2013-09-11
[MAT-199] Check for null args on construction.

gmarkall:
Unfortunately there seems to be no way to get rid of OGExpr(),
because of the one- and two-argument constructors of OGUnaryExpr
and OGBinaryExpr. However, OGExpr objects will never end up with
null args as long as OGExpr subclass constructors always ensure
that setArgs() is called if the use the default OGExpr()
constructor.

### DIFF
--- a/include/expression.hh
+++ b/include/expression.hh
@@ -95,7 +95,6 @@ class OGExpr: public OGNumeric
 class OGUnaryExpr: public OGExpr
 {
   protected:
-    OGUnaryExpr();
     OGUnaryExpr(ArgContainer* args);
     OGUnaryExpr(OGNumeric * args);
 };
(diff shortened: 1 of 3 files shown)


gmarkall:
Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/80

gmarkall:
Updated to remove noarg constructors, and stop OGExpr::_args being set as null.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/81

gmarkall:
Removed the OGUnaryExpr(arg) and OGBinaryExpr(arg1, arg2) ctors. Buildbot pass: http://devmaths-lx-1:8011/nyqwk2-testing/builds/82

stuartarchibald:
Changes made as discussed off line i.e. removing default FOO::FOO(void){}; 

Cleans things up a lot, good change set, happy to merge.

PR #16, opened 2013-09-11
[MAT-197] Implement Uncopyable

gmarkall:
This is used to make it impossible to copy OGExpr objects.

It has not been used for OGNumeric yet, because we have not yet
concluded how construction of OGTerminals should work in all
cases.

Buildbot pass: http://devmaths-lx-1:8011/nyqwk2-testing/builds/84

### DIFF
--- a/include/expression.hh
+++ b/include/expression.hh
@@ -18,6 +18,7 @@
 #include "visitor.hh"
 #include "exceptions.hh"
 #include "containers.hh"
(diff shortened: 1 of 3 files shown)


gmarkall:
Added = default and = delete for implementations.

Pass: http://devmaths-lx-1:8011/nyqwk2-testing/builds/86

stuartarchibald:
looks good.

PR #18, opened 2013-09-11
[MAT-183] Add Google Test 1.6.0 sources

gmarkall:
This enables build and test without needing Google Test to be installed. 

Also, it fixes backtraces in the librdag test code (which usually didnt work when using Google Test as a library).

CMakeParseArguments.cmake is removed as it shadows a module that is part of the CMake distribution, and was causing problems with the build.


gmarkall:
Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/92

gmarkall:
Added copyright notes.

BB pass for new revision: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/93

stuartarchibald:
0024e14 wont alter build state. happy to merge.

PR #19, opened 2013-09-11
[MAT-204] remove noCpy_ctor(), add constructor inheritance and remove and prevent default ctors.

stuartarchibald:
Addresses MAT-204 and MAT-205.
- Removes ctors containing memcpy.
- Uses constructor inheritance to expose the librdag ctors to the JNI code which then construct using the templated JNI binding functions.
- Removes the ability to create a selection of the OGArray classes with default ctors.

BB Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/104](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/104)

### DIFF
--- a/include/expression.hh
+++ b/include/expression.hh
@@ -40,7 +40,7 @@ class OGIntegerScalar;
 /*
  * Base class for absolutely everything!
  */
-class OGNumeric
+class OGNumeric: private Uncopyable
 {
   public:
     virtual ~OGNumeric();
@@ -53,7 +53,7 @@ class OGNumeric
     OGNumeric();
 };
 
-/* 
+/*
  * Base class for terminal nodes in the AST
  */
 class OGTerminal: public OGNumeric
@@ -71,7 +71,7 @@ typedef PtrVector<OGNumeric> ArgContainer;
 /**
  *  Expr type
  */
-class OGExpr: public OGNumeric, private Uncopyable
+class OGExpr: public OGNumeric
 {
   public:
     virtual ~OGExpr();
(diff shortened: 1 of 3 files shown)


include/expression.hh
```diff
@@ -606,6 +566,12 @@ template <typename T> class OGSparseMatrix: public OGArray<T>
       this->setColPtr(colPtr);
       this->setRowIdx(rowIdx);
     };
+    ~OGSparseMatrix()
+    {
+      delete(this->getData());
```
gmarkall: The sparse matrix appears to delete the data that was passed in to it on destruction. 

Is this because a copy of JVM data will be passed to the sparse matrix, or is there some other reason? (am I understanding correctly)?

gmarkall:
A few other questions/comments (not necessarily all to be resolved for this PR but seems like a good time to discuss them):
1. After you've constructed a valid OGArray object, it looks like you can get it in to an inconsistent state by calling setRows, setCols, etc on it - should these be made protected?
2. Do we want the copy constructors for terminals to take a pointer to another terminal, rather than a reference? I think the caller is most likely holding a pointer to the object they wish to copy. Secondly, it would allow all the OGNumeric classes to derive from Uncopyable, so the private default constructors on each object could be removed.
3. The copy constructors still (appear to) copy data - do we plan to make them only copy state and point to the same data at a future time? 
4. There are some very long lines (e.g. 94) - could they be split a bit please?

stuartarchibald:
BB pass [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/105](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/105) d345722

stuartarchibald:
Following discussion we decide that for now things should be kept more struct like with the idea that the classes provide "views" of data. We could see no context in which copying a view would make any sense so the most Base class extends Uncopyable which prevents copy and copy assignment ctors throughout the hierarchy.

The JNI code critical binding order needs to be nested WRT ctors and dtors to which this is now adhered.

PR #21, opened 2013-09-23
[MAT-212] Implement RTTI and copy() method

gmarkall:


stuartarchibald:
Discussion undertaken off line, copy() on OGTerminal types is used for cases where delete is called e.g. on a modified expr tree such that calling delete on the original and the modified tree won't cause a double free().

PR #22, opened 2013-09-23
[MAT-164] Implement linearisation of DAG.

gmarkall:


include/containers.hh
```diff
@@ -77,6 +73,15 @@ class PtrVector
       return c;
     }
 
+  protected:
```
stuartarchibald, line 76: Address with MAT-221

stuartarchibald:
BB PASS: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/166 docs fail only.

PR #24, opened 2013-09-23
[MAT-184] Some addition of const and virtual

gmarkall:
The original intention of this commit was to make
OGNumeric::debug_print, copy and accept const. The result was the
requirement for all the added consts.

Along the way many methods that were virtual in the superclass
were declared non-virtual in the subclass.


gmarkall:
Buildbot "pass": http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/168

PR #28, opened 2013-09-23
[MAT-225] Implement JNI handling abstraction.

gmarkall:
Commits consist of:

[MAT-225] Correct debug print
[MAT-225] Factoring things into the JVMManager compiles but fails.
[MAT-225] Implementation of missing members. Throws exception.
[MAT-225] Attempt to produce useful exception messages
[MAT-225] Native library self-test to aid debugging.
[MAT-225] Fix transposed return code check and improve error messages.
[MAT-225] Invoke NativeLibrariesSelfTest through CTest

Buildbot: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/172


stuartarchibald:
Seems good. Like the abstraction of JVMManager, saves static library level global functions.

PR #30, opened 2013-09-24
[MAT-221] Fix confusing PtrVector class hierarchy

gmarkall:
PtrVector is now the bases. OwningPtrVector and NonOwningPtrVector
implement their own copy methods that behave accordingly.

OwningPtrVector also deletes the vector elements on destruction.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/178


stuartarchibald:
Happy to merge, as discussed I think this is is cleaner as an impl.

PR #31, opened 2013-09-24
[MAT-215] Change binary MINUS into unary NEGATE

gmarkall:
As per discussion on MAT-215.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/180


stuartarchibald:
looks good.

PR #32, opened 2013-09-24
[MAT-211] add logic to complex array container to shortcut cases where the data contained is only in real space

stuartarchibald:
Adds logic to complex array container to allow programmatic shortcuts in cases where the data contained is only in real space.
e.g. 

```
ComplexArrayContainer foo = new ComplexArrayContainer(new double[][]{{1}}, new double[][]{{0}});
if(!foo.anyImaginary())
{
  // data is real only so only worry about that
}
```

BB PASS: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/181


gmarkall:
Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/181

PR #34, opened 2013-09-25
MAT-229: unit tests for C++ OGTerminals

stuartarchibald:
This PR does:
- Refactoring of `expression.hh` to remove terminal types into `terminal.hh` and the `OGNumeric` base class into `numeric.hh`
- Implements sparse matrix `toArrayOfArrays()` method
- Tests the C++ terminal types, ctors, methods etc.

BB PASS: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/187

### DIFF
--- a/src/librdag/test/check_expressions.cc
+++ b/src/librdag/test/check_expressions.cc
@@ -5,6 +5,7 @@
  */
 
 #include "expression.hh"
+#include "terminal.hh"
 #include "gtest/gtest.h"
 
 using namespace std;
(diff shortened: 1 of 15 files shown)


src/librdag/test/check_expressions.cc
```diff
@@ -201,7 +202,7 @@
 }
 
 TEST(VirtualCopyTest, RealSparseMatrix){
-  int colPtr[2] = { 0, 1 };
+  int colPtr[3] = { 0, 2, 2 };
```
gmarkall, line 205: I think this changes the matrix from being [ 1.0 0.0; 0.0 2.0 ] to [ 1.0 2.0; 0.0 0.0 ] as well as fixing the length of colPtr. I think this doesn't matter for the purpose of this test though.

gmarkall:
This looks good. I like the FakeVisitor/hasBeenVisited check, it's a good idea.

I'm happy to merge this if you are happy - do you agree that the actual matrix structure mentioned in the comment above is irrelevant, given that the structure is now valid?

src/librdag/test/check_expressions.cc
stuartarchibald, line 205: t'other way up as its CSC but yeah doesn't matter.

stuartarchibald:
Yep, agree it doesn't matter at all so long as the reads are at valid addresses.

PR #35, opened 2013-09-26
[MAT-227] Basic Register class

gmarkall:
Includes:
- Rudimentary support for invoking type-parameterised tests with
  CTest.
- OGArray::setData() no longer throws an exception for nullptr
  data. The constructors of subclasses that take data now check
  for null data instead. This is so that a Register can be
  constructed without data initially allocated.
- OGMatrix constructors are added that don't take a data pointer,
  so that the Registers can use them.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/188


gmarkall:
Not yet ready for master.

gmarkall:
Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing

stuartarchibald:
WIP merging to not cause delay upstream as Register type needed elsewhere.

PR #36, opened 2013-10-07
Mat 240

gmarkall:
Declaring PtrVector<T> to get a container of const T\* was seriously confusing. This PR redefines it so that you instead define PtrVector<const T*> to get a container of const T*.

A compile-time check enforces the types that can be used with the containers.

I can't get a full pass because of buildbot steps that have since been added, but this is what it does do: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/262

### DIFF
--- a/include/containers.hh
+++ b/include/containers.hh
@@ -25,16 +25,16 @@ class PtrVector
   public:
     PtrVector();
     virtual ~PtrVector();
-    typedef typename vector<T const *>::const_iterator citerator;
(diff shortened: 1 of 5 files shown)


gmarkall:
Should say:

> Declaring `PtrVector<T>` to get a container of const T\* was seriously confusing. This PR redefines it so that you instead define `PtrVector<const T*>` to get a container of const T*.

PR #38, opened 2013-10-08
Various changes towards increasing RDAG test coverage

gmarkall:
- Add some extra tests and fix issues discovered as a result.
- Add GTest report generation - Buildbot can now produce an HTML summary of test passes/fails
- Remove launch of individual tests through CTest - this does not work for type- and value-parameterised tests since it relies on parsing C++ code with regexes, which is impossible to do for all but the simplest cases.
- Make naming of exceptions consistent with their base class, fix their usage (i.e. avoid throwing pointers to Exceptions), and add a message to every exception.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/279


gmarkall:
Also: addition of .gitattributes that enforces sane whitespace. This hook has to be enabled manually by renamiing .git/hooks/pre-commit.sample to .git/hooks/pre-commit

PR #45, opened 2013-10-16
Mat 252 continued rebase

stuartarchibald:
Convoluted joint effort PR. Objective was to get the libjshim code unit tested via the use of a "fake-able" jvm instance.
First attempt: Intention was to fake a JVM and all associated functions by extending the jni.h structs and then function hiding. Virtuals are needed to make the fake things override and be extensible which don't go well with function hiding so this is abandoned.
Second attempt, which is committed, does the following:
- Adds a fake_jni.h which implements the parts of the jni.h header needed to get the code to build.
- Adds a CPP variable to define which jni.h to use (cmake located by default, faked if symbol defined)
- Adds the modifiermacros.h header which contains DLLEXPORT definitions for C and Fortran so that windows behaves and exports symbols as needed.

All this basically sets up code so that the libjshim library can be unit tested. Tests are then added. JVMManager is mostly covered but needs more work since rebase added extra methods.

BB PASS: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/352](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/352)

_Coverage is:_
- librdag 100%
- libjshim 21.7 %!!!!!


src/jshim/test/check_jvmmanager_fakejni.cc
```diff
@@ -0,0 +1,630 @@
...
+    {
+        return this->_env;
+    }        
+    virtual jint AttachCurrentThread(void SUPPRESS_UNUSED **penv, void SUPPRESS_UNUSED *args) { return JNI_OK; }
+};
+
+// Test bad JVM (GetEnv() env ptr returns JNI_ERR)
+TEST(JVMManagerFakeJNITest, Test_JVMManager_initialize_badversion)
+{
+    class Fake_JavaVM: public JavaVM
```
gmarkall: Calling this class Fake_JavaVM only distinguishes it from the other Fake_JavaVM by the scope it's in - could this be called something else?

stuartarchibald: Addressed in 9c9feb8
BB Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/355](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/355)

src/jshim/test/check_jvmmanager_fakejni.cc
```diff
@@ -0,0 +1,630 @@
...
+    JNIEnv * newJNIEnvPtr;
+    ASSERT_ANY_THROW(jvm_manager->getEnv((void **) &newJNIEnvPtr));
+      
+    delete jvm_manager;
+    delete jvm;
+    delete env;
+}
+
+
+// test bad getJVM()
```
gmarkall: What is bad about this getJVM?

src/jshim/test/check_jvmmanager_fakejni.cc
```diff
@@ -0,0 +1,630 @@
...
+    jvm_manager->getEnv((void **) &newJNIEnvPtr);
+    ASSERT_TRUE(newJNIEnvPtr == env);
+    
+    delete jvm_manager;
+    delete jvm;
+    delete env;
+}
+
+
+// test bad getEnv()
```
gmarkall: If I understand this test correctly, the AttachCurrentThread is the bad method - it just manifests itself when calling JVMManager::getEnv();

gmarkall:
This looks good. Aside from the inline comments, I have one or two other questions:
1. Is jni_md_linux.h dead now? Can it be removed?
2. I notice that `thread_local` has been added to the JVM Manager's `JNIEnv*`. Since the tests pass it's not causing any problems, but we might need to write some multithreaded tests in the future?
3. In general there's no need to construct a JVMManager because every method and instance variable is static. One can call `JVMManager::xxx` instead of doing it through an instance. I think this might indicate a bit of "funny" design of the JVMManager - perhaps it should have been a namespace rather than a class. It doesn't really matter for this pull request, but I wanted to mention it whilst it was on my mind.

src/jshim/test/check_jvmmanager_fakejni.cc
stuartarchibald: copy pasta

src/jshim/test/check_jvmmanager_fakejni.cc
stuartarchibald: copy pasta, will fix so as to be less confusing

stuartarchibald: actually, no, this was just a cheap way of simulating a broken getEnv()

stuartarchibald:
b5627bd addresses concerns about comments. 
BB Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/356](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/356)

stuartarchibald:
As to other Qs.
1. Yes and no. Yes it is dead. But the code needs merging+additions for other OSs adding into the `fake_jni_md.h`. This is ticket MAT-266.
2. Intent was compile only. Agree that threaded testing is needed, now if only there were a portable easy to use threading library ;-)
3. Yeah, the joys of static, this design probably needs looking at. And by probably, I mean definitely, but not top priority. This is ticket MAT-265 .

gmarkall:
1. Isn't `fake_jni_md.h` complete now? If there were something missing, wouldn't it be failing on at least one of the platforms?
   2 and 3. OK.

stuartarchibald:
1. Complete conditional on the assertion that `long long` is an `int64_t` as defined in `<cstdint>`. We can check the width at configure time. MAT-267 for checking of `long long`. More general "are we on 64bit" check needed MAT-268.

Removed the offending `jni_md_linux.h` in d75e0ca.

BB Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/357](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/357)

PR #46, opened 2013-10-17
Mat 235 rebase rebase

stuartarchibald:
As suggested this is a rebase rebase nightmare. 
PR is a WIP towards getting the binary operation dispatch system working.
It....
- adds a giant switch table and one of the two layers (the run layer, convert still needs writing) needed for actually executing binary nodes.
- Switches the type system to using primes for identification, this is in case hashing is needed later to identify common sequences.
- Actually manages to do 2+3=5 which demonstrates that in principle the code will work!

BB Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/361](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/361)
Coverage goes down on librdag mainly due to this being WIP and the minimal working case is vast.


include/exprtypeenum.h
```diff
@@ -16,34 +16,225 @@
...
+  LDIVIDE_ENUM         = 0X010FL ,
+  MTIMES_ENUM          = 0X0115L ,
+  COPY_ENUM            = 0X0119L ,
+  SELECTRESULT_ENUM    = 0X011BL ,
+  SVD_ENUM             = 0X0125L 
 
 } ExprType_t;
 
 #endif
+
```
gmarkall, line 55: Could this be replaced by "Magic numbers for terminals: primes up to 256. Magic numbers for nodes: primes between 257 and 1069"?

stuartarchibald, line 55: It could, I left the table in so that we could leave a flag of where we got to and/or write next to the number the corresponding type. Can remove if you'd prefer?

gmarkall, line 55: Nah, it';s OK.

gmarkall:
This looks OK - I haven't looked at it too hard since it's work in progress and doesn't break anything. The librdag coverage drop is not a concern since it's only new methods that are not tested, rather than the removal of any existing tests - it should be easy to rectify during some future housekeeping time.

PR #47, opened 2013-10-17
Add sparse materialisers

gmarkall:
Addition of sparse materialisers and tests. 

toArray is moved up to OGArray in librdag so that it can be accessed by diagonal and sparse matrices.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/362

Librdag coverage: 99.4% (needs another commit to get up to 100%)
jshim coverage: 20.7%

### DIFF
--- a/src/librdag/terminal.cc
+++ b/src/librdag/terminal.cc
@@ -197,6 +197,18 @@ OGArray<T>::getData() const
 }
 
 template<typename T>
+T*
+OGArray<T>::toArray() const
+{
+  T* tmp = new T[_datalen];
+  for (int i = 0; i < _datalen; i++)
+  {
+    tmp[i] = _data[i];
+  }
+  return tmp;
+}
+
+template<typename T>
 int
 OGArray<T>::getRows() const
 {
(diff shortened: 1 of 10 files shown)


gmarkall:
Rebased and added missing rdag tests.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/364

Librdag coverage: 88.6% (same as master)
jshim coverage: 20.6% (slightly lower than master, due to adding some untested functions in a set of similar untested functions)

src/librdag/terminal.cc
```diff
@@ -197,6 +197,18 @@
...
 template<typename T>
+T*
+OGArray<T>::toArray() const
+{
+  T* tmp = new T[_datalen];
+  for (int i = 0; i < _datalen; i++)
+  {
+    tmp[i] = _data[i];
+  }
+  return tmp;
```
stuartarchibald, line 205: would `memcpy (tmp, _data, sizeof(T)*_datalen );` work?

gmarkall, line 205: probably. I'll try it and do so if it does. If it does, the memcpy could also probably be used for part of toArrayOfArrays.

gmarkall:
I've added a more complicated sparse matrix and fixed the bug that it exposed. Also, `toArray` is changed to use a `memcpy`.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/368

stuartarchibald:
Happy with this. Couple of good catches enabled by data with different sizes. Memcpy is valuable as our data is flat :)

PR #48, opened 2013-10-18
Add ExecutionList creation and traversal

gmarkall:
This adds execution list creation and traversal in entrypt.

The plus node runner has been changed slightly so that if its args are not terminals, it gets the register from its arguments so it is operating on the correct data.

The tree walker has also been disabled - it produces a large amount of output, and it is becoming less relevant as we being implementing actual execution. (It was also not much of a rigorous test of anything anyway).

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/382
Librdag coverage: 81% (dropped due to not calling the walker to print out the tree)
jshim coverage: 20.6% (same as master)

### DIFF
--- a/src/java/src/test/java/com/opengamma/longdog/materialisers/TestPlusMaterialise.java
+++ b/src/java/src/test/java/com/opengamma/longdog/materialisers/TestPlusMaterialise.java
@@ -1,0 +1,65 @@
+/**
+ * Copyright (C) 2013 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+
+package com.opengamma.longdog.materialisers;
+
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import com.opengamma.longdog.datacontainers.OGNumeric;
+import com.opengamma.longdog.datacontainers.scalar.OGRealScalar;
+import com.opengamma.longdog.exceptions.MathsException;
+import com.opengamma.longdog.nodes.PLUS;
+import com.opengamma.longdog.testhelpers.ArraysHelpers;
+
+public class TestPlusMaterialise {
+
+  @DataProvider
+  public Object[][] dataContainer() {
+    Object[][] obj = new Object[3][2];
+
+
+    obj[0][0] = new PLUS(new OGRealScalar(2.0), new OGRealScalar(3.0));
+    obj[0][1] = new OGRealScalar(2.0 + 3.0);
+
+    obj[1][0] = new PLUS(new PLUS(new OGRealScalar(1.0), new OGRealScalar(2.0)), new OGRealScalar(3.0));
+    obj[1][1] = new OGRealScalar(1.0 + 2.0 + 3.0);
+
+    OGNumeric bigtree = new OGRealScalar(0.0);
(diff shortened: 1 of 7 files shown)


gmarkall:
This latest version contains more tests for entrypt and is based on the current master.

The debug build was made more quiet in this branch because all the output was making it impossible to use GDB.

In order to avoid the manual cleanup, I believe that we should (in a future PR) change the Register so it deletes its data on destruction, and doesn't reference count. The present implementation was intended to work with a space-reusing allocation scheme that we're not doing for now, and it's really awkward. (My current thinking is that for now a Register is a Terminal that owns its data).

http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/410
Librdag coverage: 86.6% (new code is covered)
Jshim coverage: 20.6%

src/java/src/test/java/com/opengamma/longdog/materialisers/TestPlusMaterialise.java
```diff
@@ -0,0 +1,68 @@
...
+    double[][] answer = Materialisers.toDoubleArrayOfArrays(input);
+    if (!ArraysHelpers.ArraysEquals(new double[][] {expected.getData()}, answer)) {
+      throw new MathsException("Arrays not equal");
+    }
+  }
+
+  @Test(dataProvider = "dataContainer")
+  public void materialiseToOGTerminal(OGNumeric input, OGRealScalar expected) {
+    OGRealScalar answer = (OGRealScalar) Materialisers.toOGTerminal(input);
+    if (answer.getData().length != 1) {
```
stuartarchibald: Ctor prevents this from happening anyway. See OGRealScalar.java ~l19 (malloc(sizeof(T))) and ~l25 for Catcher on bad Number ref.

gmarkall: Do  you want this check removing?

src/librdag/entrypt.cc
```diff
@@ -133,34 +134,41 @@ void Walker::talkandwalk(librdag::OGNumeric const * numeric_expr_types)
...
     Dispatcher * disp = new Dispatcher();
-    disp->dispatch(asOGExpr);
-    const RegContainer * reg = asOGExpr->getRegs();
+    
+    DEBUG_PRINT("Dispatching from entrypt\n");
+
+    for (auto it = el->begin(); it != el->end(); ++it)
+    {
+      const OGExpr* exprToDispatch = (*it)->asOGExpr();
+      if (exprToDispatch)
```
stuartarchibald: needs C++ style. "exprToDispatch!=nullptr" ?

src/java/src/test/java/com/opengamma/longdog/materialisers/TestPlusMaterialise.java
stuartarchibald: Shan't block on it.

gmarkall:
Amended commits in response to comments. New buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/411

stuartarchibald:
Thanks.
Am happy with committed.

PR #49, opened 2013-10-21
Dispatch for unary operations

gmarkall:
Dispatch for unary operations is modelled on the dispatch for binary operations.

Its use is demonstrated with the implementation of NEGATE materialisation which is tested from both Java and C++.

This PR contains some of the removal of code that prints out trees as the PR for MAT-228, also to quieten the output down now that we're beginning to execute trees rather than just looking at them.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/385
Librdag coverage: 90.3% (2% higher than current master)
jshim coverage: 20.6% (equal to current master)

### DIFF
--- a/include/dispatch.hh
+++ b/include/dispatch.hh
@@ -132,19 +132,9 @@ public:
     // impl convert and run for type OGIntegerScalar
   };
 
-  virtual T run(RegContainer SUPPRESS_UNUSED * reg, OGRealMatrix const SUPPRESS_UNUSED * arg) const
-  {
-    // impl convert and run for type OGRealMatrix
-  };
-
   virtual T run(RegContainer SUPPRESS_UNUSED * reg, OGLogicalMatrix const SUPPRESS_UNUSED * arg) const
   {
     // impl convert and run for type OGLogicalMatrix
-  };
-
-  virtual T run(RegContainer SUPPRESS_UNUSED * reg, OGComplexMatrix const SUPPRESS_UNUSED * arg) const
-  {
-    // impl convert and run for type OGComplexMatrix
   };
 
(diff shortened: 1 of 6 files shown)


include/dispatch.hh
```diff
@@ -53,9 +54,120 @@ public:
...
+  {
+    // impl convert and run for type OGComplexScalar
+  };
+
+  virtual T run(RegContainer SUPPRESS_UNUSED * reg, OGIntegerScalar const SUPPRESS_UNUSED * arg) const
+  {
+    // impl convert and run for type OGIntegerScalar
+  };
+
+  virtual T run(RegContainer SUPPRESS_UNUSED * reg, OGRealMatrix const SUPPRESS_UNUSED * arg) const
```
stuartarchibald: should this be pure virtual as in the bin-op case such that implementers of the interface are required to provide "backstop/default" implementations so that if the operation is valid something is guaranteed to run?

gmarkall: Yes, good point. Will make this change.

stuartarchibald: Think same applies to the OGComplexMatrix variant.

gmarkall: I agree.

gmarkall:
Just added the Java-side test that I previously forgot to commit.

Buildbot: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/386

gmarkall:
The buildbot pass for the latest push is here: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/390

stuartarchibald:
Code matches dispatch logic for binary ops. Happy to merge as WIP. Tickets issued to address other known problems (e.g. large trees running from java hanging most likely due to jvm frame overflow).

PR #51, opened 2013-10-21
[MAT-251] Code done. Quitened equals method debug statements. Added tests.

stuartarchibald:
This started out as being converters and then realised the testing was impossibly tedious without `.equals()` and subsequent `.fuzzyequals()` methods. So this branch covers that with view of writing converters next.

This code:
- Adds numeric types for various combinations of `(+/-)inf` and `NaN` as constants.
- Bit wise equals for all terminals
- Fuzzy equals for all terminals
- Fuzzy comparison methods for arrays and singles of types `real16` and `complex16`
- Operators `==` and `!=` for doing bitwise equals across terminals
- Operator sequence `==~` and `!=~` for doing fuzzy equals across terminals. This is done via an overloaded `~` operator on terminals to return a "fuzzy equals container type" for comparison operators `==` and `!=` to then use the fuzzy equals variations while comparing.

BB Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/393](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/393)

New code covered to 100%. Librdag is at 82.9 %. No change to jshim or java.

### DIFF
--- a/src/librdag/equals.cc
+++ b/src/librdag/equals.cc
@@ -57,18 +57,14 @@ bool SingleValueFuzzyEquals(real16 val1, real16 val2, real16 maxabserror, real16
 {
 
 #ifdef __LOCALDEBUG
-#ifdef DEBUG
   DEBUG_PRINT("FuzzyEquals: Comparing %24.16f and %24.16f\n", val1,val2);
-#endif
 #endif
 
   // IEEE754 nans not comparable, their relation is considered "unordered" sec 5.7.
   if(std::isnan(val1))
   {
 #ifdef __LOCALDEBUG
-#ifdef DEBUG
   DEBUG_PRINT("FuzzyEquals: Failed as value 1 is NaN\n");
-#endif
 #endif
     return false;
   }
@@ -76,9 +72,7 @@ bool SingleValueFuzzyEquals(real16 val1, real16 val2, real16 maxabserror, real16
   if(std::isnan(val2))
   {
 #ifdef __LOCALDEBUG
-#ifdef DEBUG
   DEBUG_PRINT("FuzzyEquals: Failed as value 2 is NaN\n");
-#endif
 #endif
     return false;
   }
@@ -115,16 +109,12 @@ bool SingleValueFuzzyEquals(real16 val1, real16 val2, real16 maxabserror, real16
   if(maxabserror>std::fabs(diff))
   {
 #ifdef __LOCALDEBUG
-#ifdef DEBUG
   DEBUG_PRINT("FuzzyEquals: Match as below diff bounds. maxabserror > diff. (%24.16f >  %24.16f)\n", maxabserror, std::abs(diff));
-#endif
 #endif
     return true;
   }
 #ifdef __LOCALDEBUG
-#ifdef DEBUG
   DEBUG_PRINT("FuzzyEquals: Failed as diff > maxabserror. (%24.16f >  %24.16f)\n", std::abs(diff), maxabserror);
-#endif
 #endif
 
   // check if they are within a relative error bound, div difference by largest of the 2
@@ -133,17 +123,13 @@ bool SingleValueFuzzyEquals(real16 val1, real16 val2, real16 maxabserror, real16
   if(maxrelerror > relerror)
   {
 #ifdef __LOCALDEBUG
-#ifdef DEBUG
   DEBUG_PRINT("FuzzyEquals: Match as maxrelerror > relerror. (%24.16f >  %24.16f)\n", maxrelerror, relerror);
-#endif
 #endif
     return true;
   };
 
 #ifdef __LOCALDEBUG
-#ifdef DEBUG
   DEBUG_PRINT("FuzzyEquals: Fail as relerror > maxrelerror. (%24.16f >  %24.16f)\n", relerror, maxrelerror);
-#endif
 #endif
 
   return false;
--- a/src/librdag/equals.cc
+++ b/src/librdag/equals.cc
@@ -83,7 +83,7 @@ bool SingleValueFuzzyEquals(real16 val1, real16 val2, real16 maxabserror, real16
     return false;
   }
 
-// deal with infs
+// deal with infs in debug mode
 #ifdef __LOCALDEBUG
 #ifdef DEBUG
   bool val1isinf = std::isinf(val1);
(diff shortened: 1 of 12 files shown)


src/librdag/equals.cc
```diff
@@ -0,0 +1,164 @@
...
+ * of the two numbers being very small in magnitude.
+ * The parameter "maxrelerror" determines the minimum threshold for "equal" in terms
+ * of the relative magnitude of the numbers. i.e. invariant of the magnitude of the numbers
+ * what is the maximum level of magnitude difference acceptable.
+ */
+bool SingleValueFuzzyEquals(real16 val1, real16 val2, real16 maxabserror, real16 maxrelerror)
+{
+
+#ifdef __LOCALDEBUG
+#ifdef DEBUG
```
gmarkall: No need to wrap DEBUG_PRINT in ifdef DEBUG since that's already contained within the DEBUG_PRINT macro

stuartarchibald: The log ends up massively polluted with these messages if they are left in. Given the size of data sets likely, I just switched them off else you get a report of literally every floating point comparison made!

src/librdag/equals.cc
```diff
@@ -0,0 +1,164 @@
...
+        DEBUG_PRINT("FuzzyEquals: Inf Branch. Success as both inf of same sign\n");
+        return true;
+      }
+    }
+
+  DEBUG_PRINT("FuzzyEquals: Inf Branch. Fail, non matching infs\n");
+    return false;
+  }
+#endif
+#else
```
gmarkall: I think this else branch is dealing with the case where they are exactly equal - the comment above looks like it's describing this as dealing with infs though.

stuartarchibald: True. I guess the val1==val2=={+/-}0 condition below can be merged in too.

stuartarchibald: Fix in a9c0672
BB Pass [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/394](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/394)

src/librdag/terminal.cc
```diff
@@ -79,6 +141,37 @@
...
 
+template<typename T>
+bool
+OGScalar<T>::equals(const OGTerminal * other) const
+{
+  if(this->getType()!=other->getType())
+  {
+    return false;
+  }
+  if(static_cast<const OGScalar *>(other)->getValue()!=this->_value)
```
gmarkall, line 152: Is this somehow being inferred as a static cast to const OGScalar<T> \* ?

src/librdag/terminal.cc
```diff
@@ -266,6 +359,52 @@
   _datalen = datalen;
 }
 
+template<typename T>
+bool
+OGArray<T>::fundamentalsEqual(const OGTerminal * other) const
```
gmarkall, line 364: Note to self: reviewed up to here.

src/librdag/terminal.cc
gmarkall, line 152: Argh, again Github swallowed my template parameters! (The question looks ridiculous as written)

I meant to ask: is that being cast to `const OGScalar<T>*`?

I agree with your remark yesterday that using a static cast in this case is safe since we have used our own RTTI to ensure the type. This will stay safe as long as we don't use multiple inheritance where an OGScalar is not the first parent (since then the pointer will not be pointed at the OGScalar's fields). I think that this restriction is fine.

stuartarchibald, line 152: Yes it does eat code!

I also think this restriction is fine. Should the need for the case above occur, the change here would be simple, so it's not much of a lock-in/technical debt.

src/librdag/test/check_equals.cc
```diff
@@ -0,0 +1,540 @@
...
+}
+
+TEST(EqualsTest, SingleValueFuzzyEqualsComplex16) {
+  ASSERT_TRUE(SingleValueFuzzyEquals({1,10},{1,10}));
+  ASSERT_FALSE(SingleValueFuzzyEquals({1,10},{5,10}));
+  ASSERT_FALSE(SingleValueFuzzyEquals({1,10},{1,50}));
+}
+
+
+TEST(EqualsTest, ArrayBitEquals_real16) {
```
gmarkall, line 69: Some repetition can be eliminated in the EqualsTest by using fixtures.

I think it's not crucial to this PR, but something that could be done in a "fix-it" week - there are various other ways in which our tests could be improved/shortened.

src/librdag/test/check_equals.cc
```diff
@@ -0,0 +1,540 @@
...
+
+  delete [] data1;
+  delete [] data2;
+  delete [] data3;
+
+}
+
+
+TEST(EqualsTest, OGRealScalar) {
+
```
gmarkall, line 151: The EqualsTest for scalars can (Ifi've read correctly and all the conditions are the same) be rewritten as a value-parameterised test so the test code is written only once.

Again, something for a "fix-it" session?

src/librdag/equals.cc
gmarkall: I think you're talking about __LOCALDEBUG there? DEBUG_PRINT is:

```
#ifdef DEBUG
#define DEBUG_PRINT(...) fprintf(stderr, __VA_ARGS__)
#else
#define DEBUG_PRINT(...) do {} while(0)
#endif
```

so the `#ifdef DEBUG` here is redundant.

gmarkall:
This looks very good.

I see the point of `#ifdef __LOCALDEBUG` but not convinced `#ifdef DEBUG` is necessary. 

I don't think the tests need to be altered in any way at this point, unless you wanted to make any changes before this gets merged.

stuartarchibald:
Agree with comments regarding fixtures. I'm ok with things as-is in view of more pressing issues. Tickets noted as JIRA MAT-273 (test refactor) MAT-274 (quieten DEBUG_PRINT in general).

src/librdag/equals.cc
stuartarchibald: ah yes, good catch. Fix 457148e.
BB Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/395](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/395)

PR #52, opened 2013-10-22
Mat-263 Add converters

stuartarchibald:
This PR adds converters from all possible types to their possible dense matrix counterparts.
- Adds ConvertTo class
- Adds test coverage for converter class
- Makes registers memset to 0 all their allocated data space, this prevents forgetting to init to 0.
- Fixes a couple of minor bugs in the debug_print() implementations for dense matrices (data access pattern incorrect)

BB Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/406](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/406)
Coverage: New code covered to 100%. Librdag is at 84.5% (was 82.9 %), jshim and java, no change.

### DIFF
--- a/include/convertto.hh
+++ b/include/convertto.hh
@@ -34,7 +34,6 @@ class ConvertTo
     ConvertTo();
     OGRealMatrixRegister * convertToOGRealMatrix(OGRealScalar const * thing) const;
     OGRealMatrixRegister * convertToOGRealMatrix(OGRealDiagonalMatrix const * thing) const;
-    OGRealMatrixRegister * convertToOGRealMatrix(OGLogicalMatrix const * thing) const;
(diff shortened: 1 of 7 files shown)


include/convertto.hh
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Please see distribution for license.
+ */
+
+#ifndef _CONVERTTO_HH
+#define _CONVERTTO_HH
+#include "terminal.hh"
+#include "register.hh"
+
+// defines permissable type conversions
```
gmarkall, line 12: These are permissible (i.e. sensible), but not yet all implemented, if I understand correctly. Real/Complex dense are the presently available conversions?

stuartarchibald, line 12: Yes. It's similar to the DOGMA v1 "possible conversions" matrix.

PR #53, opened 2013-10-23
Generate runners automatically

gmarkall:
In the interests of trying to stick closely to master, I'm suggesting merging this first iteration of the generator that generates the runners.

First, the runners are outlined from the dispatcher in order to circumscribe the scope of what is being replaced.

Next, these files are replaced by the generator. The generator currently only contains just enough functionality to replace these files. Future additions will include the generation of more variations of functions for more nodes, and generation of parts of the dispatcher.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/416
Librdag coverage: 85.2%
Jshim coverage unchanged.


src/librdag/templates.py
```diff
@@ -0,0 +1,119 @@
...
+# Binary runner
+
+binary_runner_class_definition = """\
+class %(nodename)sRunner: public DispatchVoidBinaryOp, private Uncopyable
+{
+  public:
+    using DispatchBinaryOp<void>::run;
+    virtual void run(RegContainer* reg0, const OGComplexMatrix* arg0, const OGComplexMatrix* arg1) const override;
+    virtual void run(RegContainer* reg0, const OGRealMatrix*    arg0, const OGRealMatrix*    arg1) const override;
+    virtual void run(RegContainer* reg0, const OGRealScalar*    arg0, const OGRealScalar*    arg1) const override;
```
stuartarchibald, line 59: I'm writing this here but it's linked to the defns above. There's no pure virtual run binding for `Pair<OGRealScalar,OGRealScalar>` should we enforce it? Similarly, should the complex variant also be present and forced?

PR #57, opened 2013-10-25
Mat 276

stuartarchibald:
This is a rabbit hole PR. Was supposed to be adding code to make dispatch work by type conversion, however the need to test whether two data containers are mathematically equal, along with other general properties makes up most of the PR.
- Added converters for missing types
- Added isReal() isComplex() isScalar() isMatrix()
- Made it a requirement that dispatch returns a pointer type, this is to handle `T ret = foo(); return ret;` when `ret` is `void`
- Fixed a couple of minor but catastrophic bugs in the Materialise.toOGTerminal() code in the JNI shim.

BB PASS [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/446](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/446)
Coverage: librdag coverge: 88.5% (was 86.9%), jshim down 0.2% to 20.4% (added a couple of `toString()` methods which caused that!).

### DIFF
--- a/include/numeric.hh
+++ b/include/numeric.hh
@@ -24,6 +24,9 @@ class OGRealDiagonalMatrix;
 class OGComplexDiagonalMatrix;
 class OGRealSparseMatrix;
 class OGComplexSparseMatrix;
+class OGOwningRealMatrix;
+class OGOwningComplexMatrix;
+class ConvertTo;
 class Visitor;
 class OGExpr;
 class COPY;
@@ -31,6 +34,7 @@ class PLUS;
 class NEGATE;
 class SVD;
 class SELECTRESULT;
+
 
 /*
  * Base class for absolutely everything!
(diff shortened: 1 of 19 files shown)


include/numeric.hh
```diff
@@ -24,6 +24,9 @@ class OGRealDiagonalMatrix;
 class OGComplexDiagonalMatrix;
 class OGRealSparseMatrix;
 class OGComplexSparseMatrix;
+class OGOwningRealMatrix;
+class OGOwningComplexMatrix;
```
gmarkall, line 28: why are these owning types forward-declared here?

gmarkall, line 28: I see they are needed in terminal.hh - is that where they should be?

stuartarchibald, line 28: `convertto.hh` needs `terminal.hh` and vice versa, so to allow the cross ref the shared types are in `numeric.hh`, perhaps this is something that needs more looking at?

PR #59, opened 2013-11-07
Mat 278 squash rebase

gmarkall:
This generates the dispatcher in a similar manner to the way the runners are generated.

The manually-coded dispatch header and source files are now removed.

Once this is approved, I will squash before merging.

Buildbot pass: to follow.


gmarkall:
Will add the following logic:
- Don't free things by accident when not converting arguments, as is presently done.
- Convert to the correct type: if either arg is complex use complex converters
- If an arg is already the right type (i.e. a RealMatrix in the real case or ComplexMatrix in the complex case) then don't convert it.

gmarkall:
Also, building in parallel causes the generator to be run as many times as there are targets. This might cause spurious build failures, so I'm going to see if I can get this fixed in this PR too.

gmarkall:
I believe I now have the logic implemented as above and have also fixed the multiple generator invocation for parallel builds. I think this is ready, if you would like to check it.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/486

gmarkall:
New BB pass now dispatcher throws rdag_error when unknown type is encountered: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/487

stuartarchibald:
Addition commits were consequence of off line discussion.

Excellent. Kills off a load of manual coding. Happy to merge.

PR #60, opened 2013-11-08
Additional code generation

gmarkall:
Sufficient code generation for adding new nodes and their implementations just by adding them to the list of nodes in generator.py

Expression.cc/hh and numeric.cc/hh are now generated. Theres is some non-autogenerated code, partly because it is once-only code, and some nodes are not standard enough to obviously use the code generation we have right now (e.g. COPY, SELECTRESULT, SVD). 

This is demonstrated with the addition of the TIMES node - I have not gone ahead and added lots of nodes, preferring to discuss/get feedback on this PR first.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/490


src/librdag/dispatchtemplates.py
```diff
@@ -4,130 +4,6 @@
...
-
-# Runners header file
-
-runners_header = """\
-/**
- * Copyright (C) 2013 - present by OpenGamma Inc. and the OpenGamma group of companies
- *
- * Please see distribution for license.
- */
-
```
stuartarchibald, line 19: Do you think a comment saying "autogenned code" or similar would be a good idea just for absolute clarity ? Perhaps also mentioning the generator?

src/librdag/exprtemplates.py
```diff
@@ -0,0 +1,232 @@
...
+
+# Expressions header
+
+expression_hh = """\
+/**
+ * Copyright (C) 2013 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+
```
stuartarchibald, line 17: same comment on autogenned code as above

src/librdag/dispatchtemplates.py
gmarkall, line 19: Yes, that is a good idea. Will add it.

src/librdag/exprtemplates.py
gmarkall, line 17: Noted.

stuartarchibald:
Pending addition of commentary regarding autogenned nature of C++, happy to merge following new BB pass.

gmarkall:
Added note on autogeneration to all templates (not just the ones affected here).

New buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/492

PR #62, opened 2014-02-24
Use non-critical JNI calls

gmarkall:
Non-critical JNI calls for getting and releasing array elements
are swapped for the critical calls. On release, the native copy
of data is copied back to Java and freed. This is towards
ensuring thread-safety of the codebase.

Since the non-critical calls require the pointer to native data,
a bug in the initialisation of scalars that caused the data ref
to get lost was exposed. This is fixed by initialising a parent
OGScalar of a Java OGScalar to a default value, grabbing the
data ref, then using the newly-obtained ref to update the parent
scalar's value.

An adaptor for calling the JNI get/release functions using a
templated interface is added, to fit the JNI functions into the
existing structure of `bindPrimitiveArrayData` et al. functions.

Because a jint on Windows is a long int, as opposed to the int
used on Linux/OS X, extra casts are required in the binding, but
these will disappear once MAT-180 (use `ogint_t` for all integers)
is resolved.

The extra JNI methods that are now used are added to the fake JVM
for testing purposes.

Much output was produced, that would make it difficult to see any
errors coming from `-Xcheck:jni`. The majority of the output is
removed or suppressed by this PR.

`-Xcheck:jni` is also added on Windows.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/538
Coverage: librdag 92.3%, jshim 20.6% (essentially unchanged)


gmarkall:
Added check for nulls from `GetDoubleArrayElements` and `GetIntArrayElements`. 

New buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/539

stuartarchibald:
This looks good. Good catch on the ctor bug. I'm happy with this now the defensive checks are in. ... Merging.

PR #63, opened 2014-02-28
Mat 298

stuartarchibald:
This is a pre-PR, PR, purely for the purposes of discussing the implementation inline.
Main discussion point is whether: [https://github.com/OpenGamma/nyqwk2/blob/859f6063f66286160e64b47b8ac7121ca7c96599/src/librdag/test/nodes/check_norm2.cc](https://github.com/OpenGamma/nyqwk2/blob/859f6063f66286160e64b47b8ac7121ca7c96599/src/librdag/test/nodes/check_norm2.cc)
is good enough and whether some (needed?!) member exposure in the present hierarchy is worth it for the gains in testing ease in the above.

Similarly, comments elsewhere appreciated.

Errors on windows and mac compilers still need fixing up.


src/librdag/expressionbase.cc
```diff
@@ -28,13 +28,28 @@
...
 
+OGExpr::OGExpr() {}
+
 OGExpr::~OGExpr()
 {
   delete this->_args;
   delete this->_regs;
 }
 
-const ArgContainer*
```
gmarkall, line 37: getArgs and getRegs have lost their constness, but do they need to? I think that although you might change which ArgContainer/RegContainer an expression points to, the container that they're pointing to doesn't change - is that claim true?

stuartarchibald:
Thanks for the comments. Closing this, managed to template and macro my way out of it as suggested :)

PR #64, opened 2014-03-03
Mat 298

stuartarchibald:
This PR is decidedly long, for which I apologise. The aim was simply to get in a call to LAPACK and BLAS to demonstrate that we have access to the libraries and the calls work.
- First the LAPACK build was adjusted to yield a file containing exportable targets (`LAPACK_EXPORT`) https://github.com/OpenGamma/OG-Lapack/pull/3.
- The CMake build was updated to deal with MULTI_LINK libraries (the instruction set based libraries).
- The CMake build now imports the `LAPACK_EXPORT`s file to access the LAPACK libraries. The CMake was further updated to ensure inclusion of these targets in the built jar and in run time locations for OS's with weak library resolution capabilities.
- Once LAPACK was added the node `NORM2` (as in the Euclidean norm) was inserted into the python generator and additional custom implementation code was added. This code calls functions from BLAS and LAPACK to compute the norm, thus correct test results indicate correct linkage and calling. This code also required the implementation of an `isVector()` function for indicating vector contexts on matrices.
- The java code got a corresponding NORM2 node and unit tests, the `FuzzyEquals` class was added to aid in testing these.
- The C++ code got a nodes test subdir which contains some classes and macros to make testing nodes easier (i.e. https://github.com/OpenGamma/nyqwk2/compare/master...MAT-298-cleanup#diff-eaf3b791f5dfc65be4084d27852a71d1R24).
- GTest is now built as a shared library so it can be relocated.

BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/569](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/569)
Coverage: librdag 91.9%, jshim 20.5% (slight regression, [MAT-305](http://jira.opengamma.com/browse/MAT-305) to address, no point in hand coding it!)

If it is felt that [MAT-305](http://jira.opengamma.com/browse/MAT-305) should go in here, I'm happy to comply. Just felt it was already getting huge as a change set.


gmarkall:
I agree that some more coverage checking should be automated - for example, I think that here `CHECK_CAST(NORM2)` could be added to `check_rtti.c` to add back a little of the coverage of numeric_base that has been lost. And also for some testing of the copy functions. However, I don't think it's worth making a change for this one case for now, and we might as well save it for MAT-305.

gmarkall:
Please ignore my comments on individual diffs - they do appear to be outdated by later commits

gmarkall:
This looks great. There is the coverage issue that needs to be worked out, but I think we agree that it's a separate issue and not worth piling into this PR. 
Merging.

PR #66, opened 2014-03-10
Propagation of exceptions into Java (and other small fixes.

gmarkall:
On the RDAG side, exceptions with a backtrace are added:

This required exceptions to be split into a separate namespace,
dogma_exceptions, since it contains code common to librdag and
convert.

The exceptions use GNU execinfo to construct a backtrace and
record this in BacktraceElement objects, one for each frame.

The existing rdag_error and convert_error classes are now direct
subclasses of dogma_error.

Some testing is added to check the exception classes where
possible. Since the stack can vary from one environment to the
next, only the topmost frames are examined for their function
names (addresses and filenames can vary from run to run). In the
case of assignment and copy, comparisons are made between the
original and new object to avoid this inconsistency between
machines/runs being a problem.

On the JShim/Java side:

The C++ exceptions are caught by try-catch blocks in the highest
level JNI methods, and convert any caught exception to the
appropriate corresponding Java exception.

The MathsExceptionNative class inserts the backtrace passed to its
constructor by the JNI code into the Java stack trace. This is the
base for exceptions in native code. Conversion or computation
exceptions actuall instantiate MathsExceptionNativeConversion or
MathsExceptionNativeComputation.

This does not address release of resources in the presence of
exceptions - i.e leaks are still possible. Exception safety against
memory leaks remains to be addressed in future.

Other small fixes include:

Addition of valgrind suppressions for apparent leaks in dlopen.
Adding missing checks of the return code from Lapack in the norm runner, and fixes to the names of the called functions in the comments.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/646
Coverage: RDAG: 91.5%, jshim 19%, Java 80%.

### DIFF
--- a/include/exceptions.hh
+++ b/include/exceptions.hh
@@ -57,6 +57,10 @@ class BacktraceElement
      */
     const string getFunction() const;
   private:
+    /**
+     * Converts any non-ascii characters to ascii characters
+     */
+    string asciiOnly(string s);
     const void* _address;
     string _object_file;
     string _function;
--- a/include/exceptions.hh
+++ b/include/exceptions.hh
@@ -10,6 +10,11 @@
 #include <exception>
 #include <stdexcept>
 
+/**
+ * Maximum number of frames to retrieve for a stack trace. A somewhat arbitrary
+ * choice, but probably more than 50 stack frames implies you have worse problems
+ * than needing to know what was going on in the 51st frame.
+ */
 #define FRAMEBUF_SIZE 50
 
 using namespace std;
--- a/include/exceptions.hh
+++ b/include/exceptions.hh
@@ -10,28 +10,111 @@
 #include <exception>
 #include <stdexcept>
 
+#define FRAMEBUF_SIZE 50
+
 using namespace std;
+
+namespace dogma_exceptions {
+
+/**
+ * Details of one stack frame in a backtrace.
+ */
+class BacktraceElement
+{
+  public:
+    /**
+     * Default constructor required for new[]
+     */
+    BacktraceElement();
+    /**
+     * Copy constructor
+     *
+     * @param other the BacktraceElement to copy
+     */
+    BacktraceElement(const BacktraceElement& other);
+    /**
+     * Constructor that is usually used for constructing a new BacktraceElement.
+     *
+     * @param backtrace_symbol a single description of a frame from backtrace_symbols()
+     * @param address the address of the instruction pointer for the frame
+     */
+    BacktraceElement(const string& backtrace_symbol, const void* address);
+    /**
+     * Get the address of the instruction pointer for the frame.
+     */
+    const void* getAddress() const;
+    /**
+     * Get the name and path of the object file that this frame is inside.
+     */
+    const string getObjectFile() const;
+    /**
+     * Get the name of the function that this frame is inside.
+     */
+    const string getFunction() const;
+  private:
+    const void* _address;
+    string _object_file;
+    string _function;
+};
+
+/**
+ * Any exception thrown during the execution of the DOGMA libraries
+ */
+class dogma_error: public runtime_error
+{
+  public:
+    /**
+     * The usual constructor for building a dogma_error.
+     *
+     * @param what Description of the exception, in a form suitable for printing to the user as the
+     *             stack trace in Java.
+     */
+    dogma_error(const string& what);
+    /**
+     * Copy constructor
+     *
+     * @param other the dogma_error to copy
+     */
+    dogma_error(const dogma_error& other);
+    /**
+     * Destructor. Deletes all the BacktraceElements that this exception holds.
+     */
+    virtual ~dogma_error();
+    /**
+     * Get an array of all frames in the backtrace. The array is owned (and deleted) by the dogma_error.
+     */
+    const BacktraceElement* getBacktrace() const;
+    /**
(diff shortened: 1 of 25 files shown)


gmarkall:
New buildbot pass after fixing my erroneous push: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/649

stuartarchibald:
Thanks for patching those minor issues. I'll re-review in light of them :)

gmarkall:
I still have the more major fixes to push - do you want to wait until I've done that to save going round 3 times?

stuartarchibald:
This is really good in terms of execution of ideas, dead looking forward to wiring it in and showing it off :). I'm happy to wait for further fixes, this is not a blocker for me, I think we are both waiting on [MAT-322](http://jira.opengamma.com/browse/MAT-322), which is in turn blocked by [MAT-323](http://jira.opengamma.com/browse/MAT-323) which I'm working on at the minute. MAT-322 is actually implemented on it's branch, it's just the testing part that needs MAT-323 as it needs the concept "materialise to persisting terminal" implemented.

gmarkall:
From our conversation yesterday I think this is the list of things that remains:
- Check results of JNI calls that are not checked already
- Remove commented code in exception test
- use `ensureLocalCapacity` to check frame space before making JNI calls
- Check encoding of strings passed to `NewStringUTF`

gmarkall:
So, I think I've addressed:
- The results of JNI calls are checked. Mostly this is accomplished by checking the return value - in the cases where the return value is checked for nullptr, the JNI specification states that null is returned if and only if a Java exception has been set by the method. In the other cases, (e.g. for `SetObjectArrayElement`) we call `checkEx` after the call. If any exception is detected, we return back to Java as quickly as possible so it can deal with the exception appropriately. Prior to doing so, we make an attempt to let the user know that something has gone wrong with a message to stderr, which is about all is in our power to do in that situation.
- Removed the commented check on the size of the frame in the exception tests.
- Used `EnsureLocalCapacity` to make sure that exception handling code has enough local references available. If there are not enough, we return to Java as fast as possible.
- The strings that are passed to `NewStringUTF` originate from the symbols in the ELF file - backtrace symbols returns an array of pointers to the symbol table from the file, as far as I can ascertain. This is about as sure as I can be that the strings passed to `NewStringUTF` are encoded with a UTF-compatible encoding (which includes ASCII).

gmarkall:
Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/671
Coverage: RDAG 92.4% (same), jshim 65.7% (drop due to exception translation code that is not tested)
Java: 82% (same)

The testing of the exception translation can hopefully be done more easily after the design of the JVMManager has been adjusted. The test of exception propagation from Java will come in a future pull request using invalid matrix multiplications - this one has been going on a long time.

include/exceptions.hh
```diff
@@ -10,28 +10,116 @@
 #include <exception>
 #include <stdexcept>
 
+/**
+ * Maximum number of frames to retrieve for a stack trace. A somewhat arbitrary
+ * choice, but probably more than 50 stack frames implies you have worse problems
+ * than needing to know what was going on in the 51st frame.
```
stuartarchibald, line 16: Best comment I've seen in quite some time :)

stuartarchibald:
I think it is almost there, which I'm sure will be a relief. Couple of comments:
- `checkEx()` probably should be a `if(env->ExceptionCheck()==JNI_TRUE) return nullptr;` as the exception from java would be a valid one and therefore quick return so the JVM can complain about it is needed.
- `&0x7F` the ASCII bytes to make sure only the low bits are set?

src/jshim/jshim.cc
```diff
@@ -31,23 +211,38 @@
...
+  catch (convert_error e)
+  {
+    convertExceptionJava(env, e);
+    return nullptr;
+  }
+  catch (rdag_error e)
+  {
+    rdagExceptionJava(env, e);
+    return nullptr;
+  }
```
stuartarchibald, line 243: We need a more general exception handler either via `set_terminate()` or as a catch all here. This is not going in this unit of work as it is already enormous! Ticket [MAT-332](http://jira.opengamma.com/browse/MAT-332) is for the catch all and [MAT-182](http://jira.opengamma.com/browse/MAT-182), which has been around a while, is for a termination handler.

src/librdag/test/check_exceptions.cc
```diff
@@ -0,0 +1,144 @@
...
+
+TEST_F(DogmaErrorTest, Assign)
+{
+  dogma_error e2 = *_error;
+
+  // Check what is the same
+  EXPECT_STREQ(_error->what(), e2.what());
+
+  // Check for distinct but equal backtraces
+  ASSERT_EQ(_error->getTraceSize(), e2.getTraceSize());
```
stuartarchibald, line 135: Any chance the base pointers could be checked for unique addresses please?

gmarkall:
Remaining three points addressed. Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/673

stuartarchibald:
Thanks for fixing up those few things. Great bit of code :) A battle like Thermopylae, what an apt ticket. Really happy to merge this.

PR #67, opened 2014-03-11
[MAT-323] add owning copiers

stuartarchibald:
This PR addresses MAT-323, the creation of owning copies. The purpose of these copies is so that at any point in execution a terminal can be extracted to a clean data-owning copy by duplication. This is so that part or all of an execution tree can be deleted, manipulated or indeed just mangled with no bearing on the "liveness" of an output terminal.
- Adds `createOwningCopy()` which creates an exact data-owning copy of the terminal.
- Adds `createComplexOwningCopy()` which creates a complex type variant of a terminal with data-owning properties.
- Adds the convenience ctor `OGOwningComplexMatrix(real16 * data, int rows, int cols);` due to its prolific use potential.

BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/659](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/659).
Coverage: librdag 92.4%, jshim 20.5% (slight improvement in librdag)

### DIFF
--- a/include/terminal.hh
+++ b/include/terminal.hh
@@ -150,8 +150,8 @@ class OGRealScalar: public OGScalar<real16>
     virtual ExprType_t getType() const override;
     virtual OGOwningRealMatrix * asFullOGRealMatrix() const override;
     virtual OGOwningComplexMatrix * asFullOGComplexMatrix() const override;
-    virtual OGTerminal * createOwningCopy() const;
-    virtual OGTerminal * createComplexOwningCopy() const;
+    virtual OGTerminal * createOwningCopy() const override;
+    virtual OGTerminal * createComplexOwningCopy() const override;
 };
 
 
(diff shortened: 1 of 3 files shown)


include/terminal.hh
```diff
@@ -142,6 +150,8 @@ class OGRealScalar: public OGScalar<real16>
     virtual ExprType_t getType() const override;
     virtual OGOwningRealMatrix * asFullOGRealMatrix() const override;
     virtual OGOwningComplexMatrix * asFullOGComplexMatrix() const override;
+    virtual OGTerminal * createOwningCopy() const;
```
gmarkall: Can you add "override" to each overridden method so the compiler can catch errors where we are not overriding something where we thought we were (we have had a quite a few of that type in the past).

stuartarchibald: The compiler would whine due to their pure virtual nature, though from a human perspective I completely agree it's a good idea and have added them, well spotted.
Patched in: https://github.com/OpenGamma/nyqwk2/commit/70512e87adbaf1988b4e692af68fbbde4199431a
BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/660](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/660)

stuartarchibald:
Patch a0200bf adds in type check when `%` (`mathequals()`) operator used for comparison and adds in reinterpret cast check on data pointer to prevent regressions should the API change.
BB Multi-Pass:[http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/661](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/661)

gmarkall:
This looks great - although the copying to complex versions was a bit fiddly you seem to have made it into a nice solution.

PR #68, opened 2014-03-11
Mat 317

stuartarchibald:
This PR is fairly heavy going I'm afraid. MAT-317 is to increase the coverage of the JNI layer, which as we know is horribly complicated to test. I feel the outcome of this code is that we now have a large number of tests covering a lot of behaviour, which is good, however it emphasises that perhaps the design of the JVMManager class is wrong. The static nature of the methods in the JVMManager (for sound design reasons at the time) make it very hard to test as the underlying JNI calls have to be mocked opposed to mocking the behaviour of the JVMManager methods via class extension and overriding virtuals.

This PR:
- Adds a load of unit tests for the JNI layer (largely terminals and data binding).
- Sorts out problems with windows symbol exports.
- Removes duplicated binding code.
- Fixes a number of erroneous `delete` calls.
- Makes methods that are not implemented throw opposed to just run debug statements (done with view that there's work waiting to make it to mainline that handles exceptions).
- Adds more aggressive checking of JNI calls and state.

I realise that the jshim test code is not that eloquent and has some duplication. However, as a redesign is on the books, I think the effort of neater testing with an mock JVMManager would be best done at that time.

BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/662](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/662)
Coverage: librdag 92.4%, jshim 71.0% (massive improvement in jshim)

### DIFF
--- a/src/jshim/jdispatch.cc
+++ b/src/jshim/jdispatch.cc
@@ -146,7 +146,7 @@ void DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<complex16> const 
 }
 void DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<real16> const SUPPRESS_UNUSED  *thing)
 {
-    throw std::runtime_error("DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<real16>)");
+    throw std::runtime_error("DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<int>)");
 }
 void DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<int> const SUPPRESS_UNUSED *thing)
 {
--- a/src/jshim/jdispatch.cc
+++ b/src/jshim/jdispatch.cc
@@ -28,9 +28,9 @@ DispatchToReal16ArrayOfArrays::~DispatchToReal16ArrayOfArrays()
 
         for(int i=0; i<this->getRows(); i++)
         {
-            delete arr[i];
+            delete[] arr[i];
         }
-        delete arr;
+        delete[] arr;
     }
 }
 
@@ -127,12 +127,11 @@ DispatchToComplex16ArrayOfArrays::~DispatchToComplex16ArrayOfArrays()
     complex16 ** arr = this->getData();
     if(arr)
     {
-
         for(int i=0; i<this->getRows(); i++)
         {
-            delete arr[i];
+            delete[] arr[i];
         }
-        delete arr;
+        delete[] arr;
     }
 }
 void DispatchToComplex16ArrayOfArrays::visit(librdag::OGExpr const SUPPRESS_UNUSED *thing)
--- a/src/jshim/jdispatch.cc
+++ b/src/jshim/jdispatch.cc
@@ -36,7 +36,7 @@ DispatchToReal16ArrayOfArrays::~DispatchToReal16ArrayOfArrays()
 
 void DispatchToReal16ArrayOfArrays::visit(librdag::OGExpr const SUPPRESS_UNUSED *thing)
 {
-    throw std::runtime_error("DispatchToReal16ArrayOfArrays::visit(librdag::OGExpr)");
+    throw convert_error("DispatchToReal16ArrayOfArrays::visit(librdag::OGExpr)");
 }
 
 void DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<real16> const  *thing)
@@ -48,11 +48,11 @@ void DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<real16> const  *thin
 
 void DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<complex16> const SUPPRESS_UNUSED  *thing)
 {
-    throw std::runtime_error("DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<complex16>)");
+    throw convert_error("DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<complex16>)");
 }
 void DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<int> const SUPPRESS_UNUSED  *thing)
 {
-    throw std::runtime_error("DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<int>)");
+    throw convert_error("DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<int>)");
 }
 
 void DispatchToReal16ArrayOfArrays::visit(librdag::OGMatrix<real16> const *thing)
@@ -64,7 +64,7 @@ void DispatchToReal16ArrayOfArrays::visit(librdag::OGMatrix<real16> const *thing
 
 void DispatchToReal16ArrayOfArrays::visit(librdag::OGMatrix<complex16> const SUPPRESS_UNUSED *thing)
 {
-    throw std::runtime_error("DispatchToReal16ArrayOfArrays::visit(librdag::OGMatrix<complex16>)");
+    throw convert_error("DispatchToReal16ArrayOfArrays::visit(librdag::OGMatrix<complex16>)");
 }
 void DispatchToReal16ArrayOfArrays::visit(librdag::OGDiagonalMatrix<real16> const *thing)
 {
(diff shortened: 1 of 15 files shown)


stuartarchibald:
I rewrote history as the commit log was ridiculous, new BB Mulit-Pass:
[http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/663](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/663)

src/jshim/jdispatch.cc
```diff
@@ -147,7 +146,7 @@ void DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<complex16> const
 }
 void DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<real16> const SUPPRESS_UNUSED  *thing)
 {
-    throw std::runtime_error("DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<real16>)");
+    throw std::runtime_error("DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<int>)");
```
gmarkall: isn't this in `DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<real16>` ?

src/jshim/test/check_jbindings.cc
```diff
@@ -0,0 +1,253 @@
...
+    ASSERT_ANY_THROW(clazz->run(obj,meth));
+    delete obj;
+    delete clazz;
+}
+
+TEST(JBindings, Test_bindPrimitiveArrayData_bad_env)
+{
+    class Fake_JavaVM_bad_env: public Fake_JavaVM
+    {
+      virtual jint AttachCurrentThread(void SUPPRESS_UNUSED **penv, void SUPPRESS_UNUSED *args) override
```
gmarkall: penv is not unused

src/jshim/test/check_jbindings.cc
```diff
@@ -0,0 +1,253 @@
...
+    }
+    void setEnv(JNIEnv * env)
+    {
+        this->_env = env;
+    }
+    JNIEnv * getEnv()
+    {
+        return this->_env;
+    }
+    virtual jint AttachCurrentThread(void SUPPRESS_UNUSED **penv, void SUPPRESS_UNUSED *args)
```
gmarkall: penv is not unused

src/jshim/test/check_jbindings.cc
```diff
@@ -0,0 +1,253 @@
...
+  delete[] data;
+  delete obj;
+  delete clazz;
+}
+
+TEST(JBindings, Test_unbindPrimitiveArrayData_bad_env)
+{
+    class Fake_JavaVM_bad_env: public Fake_JavaVM
+    {
+      virtual jint AttachCurrentThread(void SUPPRESS_UNUSED **penv, void SUPPRESS_UNUSED *args) override
```
gmarkall: penv is not unused

src/jshim/test/check_jterminals.cc
```diff
@@ -0,0 +1,499 @@
...
+
+template<typename T> class Fake_JNIEnv_for_binding: public Fake_JNIEnv
+{
+  public:
+    Fake_JNIEnv_for_binding(T value):Fake_JNIEnv()
+    {
+      _value = value;
+    }
+    // fake object call returns obj
+    virtual jobject CallObjectMethod(jobject SUPPRESS_UNUSED obj, jmethodID SUPPRESS_UNUSED methodID, ...) override
```
gmarkall: obj not unused

src/jshim/test/check_jterminals.cc
```diff
@@ -0,0 +1,499 @@
...
+  public:
+    Fake_JNIEnv_for_OGMatrix_T(int rows, T * value):Fake_JNIEnv()
+    {
+      _rows = rows;
+      _value = value;
+    }
+    virtual jint CallIntMethod(jobject SUPPRESS_UNUSED obj, jmethodID SUPPRESS_UNUSED methodID, ...) {
+      return _rows;
+    }
+    virtual jobject CallObjectMethod(jobject SUPPRESS_UNUSED obj, jmethodID SUPPRESS_UNUSED methodID, ...) override
```
gmarkall: obj is not unused

src/jshim/test/check_jterminals.cc
```diff
@@ -0,0 +1,499 @@
...
+{
+  public:
+    Fake_JNIEnv_for_OGSparseMatrix_T(int rows, T * value, int * idx):Fake_JNIEnv_for_OGMatrix_T<T>(rows,value)
+    {
+      _idx = idx;
+    }
+    virtual jint CallIntMethod(jobject SUPPRESS_UNUSED obj, jmethodID SUPPRESS_UNUSED methodID, ...) {
+      return this->_rows;
+    }
+    virtual jobject CallObjectMethod(jobject SUPPRESS_UNUSED obj, jmethodID SUPPRESS_UNUSED methodID, ...) override
```
gmarkall: obj is not unused

gmarkall:
This looks excellent, and looks like a thorough test of everything that can go right/wrong in the tested parts of jshim. I only see a couple of minor issues and discussion points:
- The `Fake_JavaVM` and `Fake_JNIEnv` are replicated across each jshim test. Is there a reason for this, or can they be combined in a single location?
- `getEnv` is not a real JNI method, and looks like a convenience method for setting up tests. I note that it wasn't actually added as part of this PR, but it looks similar enough to `GetEnv` to trip us (me) up in future - can we change the name to make it more obvious what it is, or just use `GetEnv` instead?
- There are a few places where `SUPPRESS_UNUSED` is annotated on variables that are not actually unused. I started marking all the instances, but then I noticed it was mainly just `penv` and they are probably a copy/paste error. I annotated the other extraneous `SUPPRESS_UNUSED` markings that I saw though.
- Whilst looking at the tests for jbindings/jterminals, I noticed that we check for null being returned by `GetDoubleArrayElements` etc. The JNI spec states that it (and other Get<Type>ArrayElements methods) return null when there is a Java exception pending. However, we immediately throw a C++ exception that will be transformed into a Java one by our outer try/catch. I think the result of this will be that the original Java exception may be getting suppressed (I'm not 100% sure - if 2 exceptions happen in JNI, which one wins? is it the last one?).
- The dispatch functions are throwing `std::runtime_error`, and I think this is a hangover from our original (pre-defining our own exceptions) implementation. I think these can be changed to `convert_error` now.

stuartarchibald:
Thanks for reviewing...
- The replication is because I like knowing where things are due to my small mental stack :) I'll merge in a tick.
- Following a long discussion, [MAT-329](http://jira.opengamma.com/browse/MAT-329) is there to deal with the general `env` problems and to make the manager threadsafe in the process. Unfortunately it's going to create a lot of method churn.
- Will remove `SUPPRESS_UNUSED` where needed.
- I rather suspect its random undefined behaviour so I'll correct such that checkEx() is called post call.
- Yeah will fix.

stuartarchibald:
I think I've addressed 1,3,4,5 if you wouldn't mind taking a look please? And added a ticket for 2 as that's a few days work!
New BB Multi-pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/666](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/666)

src/jshim/test/check_jterminals.cc
```diff
@@ -0,0 +1,436 @@
...
+  jmethodID meth = new _jmethodID();
+  ASSERT_ANY_THROW(getIntFromVoidJMethod(meth,nullptr));
+  delete meth;
+}
+
+TEST(JTerminals, Test_getIntFromVoidJMethod_null_envptr)
+{
+  class Fake_JavaVM_bad_attach: public Fake_JavaVM
+  {
+    virtual jint AttachCurrentThread(void SUPPRESS_UNUSED **penv, void SUPPRESS_UNUSED *args) override
```
gmarkall: penv is not unused here

stuartarchibald:
Patched [https://github.com/OpenGamma/nyqwk2/pull/68#discussion_r10512423](https://github.com/OpenGamma/nyqwk2/pull/68#discussion_r10512423) in 9a51a81. 
BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/667](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/667)

gmarkall:
Apologies for my last comment! This looks great.

PR #69, opened 2014-03-12
[MAT-334] Add constructors required for Windows implementation

gmarkall:
Hopefully this should fix the missing constructor error on Windows.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/677

### DIFF
--- a/src/java/src/main/java/com/opengamma/longdog/exceptions/MathsExceptionNative.java
+++ b/src/java/src/main/java/com/opengamma/longdog/exceptions/MathsExceptionNative.java
@@ -6,6 +6,14 @@ package com.opengamma.longdog.exceptions;
 public class MathsExceptionNative extends MathsException {
 
   /**
+   * This form of the constructor is used on Windows, where no native stack trace is provided.
+   * @param msg the error message
+   */
+  public MathsExceptionNative(String msg) {
+	super(msg);
+  }
+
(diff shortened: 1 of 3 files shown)


gmarkall:
Just waiting on platform pass with this version of nyqwk2...

stuartarchibald:
I think the fix is sound but I'm fairly sure the build will fail as the symbol table that highlighted this problem is from the `MTIMES()` node which isn't in mainline yet (it's WIP on branch MAT-322 of nyqwk2). I've added https://github.com/OpenGamma/nyqwk2/tree/MAT-322_with_MAT-334 as a combo of these branches which passes here: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/678](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/678). The MAT-310 (the native integration WIP) branch of platform passes too with the above branch build [http://devmaths-lx-1:8011/builders/platform_nyqwk2-testing/builds/15](http://devmaths-lx-1:8011/builders/platform_nyqwk2-testing/builds/15), so I think we can say this is an acceptable hot-fix pending end-to-end testing that will be provided by MAT-322.

stuartarchibald:
Merging as those involved are happy that this is correct.

PR #70, opened 2014-03-18
Mat 341

stuartarchibald:
This ticket was about making it possible to instantiate the templated terminal types and removing the `*Owning*` classes in favour of just having data ownership as a property of the class.
The PR:
- Removes all the `*Owning*` classes.
- Adds the enum `DATA_ACCESS` with members `OWNER` and `VIEWER`, to reflect ownership of underlying POD or viewing of underlying POD respectively.
- Adds ctors that take a `DATA_ACCESS` parameter to all classes that extend `OGArray<T>`.
- Updates the unit tests to reflect the above.
- Adds in a load of default methods that throw as implementations for currently abstract templates, like `OGMatrix<T>`, such that they aren't abstract and can be instantiated.
- Adds tests and regression tests such that the above mention templates cannot accidentally become abstract again!

BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/713](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/713)
Coverage: librdag 92.4%, jshim 63.9% (jshim decrease due to merge of untested exception handling code from https://github.com/OpenGamma/nyqwk2/pull/66)
Java coverage: 81% (1% decrease, reason as for jshim)

### DIFF
--- a/src/librdag/terminal.cc
+++ b/src/librdag/terminal.cc
@@ -243,6 +243,48 @@ OGScalar<T>::fuzzyequals(const OGTerminal * other) const
   return true;
 }
 
+template<typename T>
+OGRealMatrix *
+OGScalar<T>::asFullOGRealMatrix() const
+{
+  throw rdag_error("Cannot asFullOGRealMatrix() template OGScalar<T> class.");
+}
+
+template<typename T>
+OGComplexMatrix  *
+OGScalar<T>::asFullOGComplexMatrix() const
+{
+  throw rdag_error("Cannot asFullOGComplexMatrix() template OGScalar<T> class.");
+}
+
+template<typename T>
+OGTerminal *
+OGScalar<T>::createOwningCopy() const
+{
+  throw rdag_error("Cannot createOwningCopy() template OGScalar<T> class.");
+}
+
+template<typename T>
+OGTerminal *
+OGScalar<T>::createComplexOwningCopy() const
+{
+  throw rdag_error("Cannot createComplexOwningCopy() template OGScalar<T> class.");
+}
+
+template<typename T>
+void
+OGScalar<T>::debug_print() const
+{
+  cout << "OGScalar<T>: " << setprecision(__DEBUG_PRECISION) << this->getValue() << endl;
+}
+
+template<typename T>
+OGNumeric *
+OGScalar<T>::copy() const
+{
+  throw rdag_error("Cannot copy() template OGScalar<T> class.");
+}
+
 
 template class OGScalar<real16>;
 template class OGScalar<complex16>;
--- a/src/librdag/terminal.cc
+++ b/src/librdag/terminal.cc
@@ -1021,7 +1021,7 @@ OGRealDiagonalMatrix::asFullOGComplexMatrix() const
 OGTerminal *
 OGRealDiagonalMatrix::createOwningCopy() const
 {
-  real16 * newdata =  new real16[this->getDatalen()]();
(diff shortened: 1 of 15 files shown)


src/librdag/terminal.cc
```diff
@@ -921,25 +1021,17 @@ bool OGTerminal::operator!=(const detail::FuzzyCompareOGTerminalContainer& thing
 OGTerminal *
 OGRealDiagonalMatrix::createOwningCopy() const
 {
-  real16 * newdata =  new real16[this->getDatalen()];
+  real16 * newdata =  new real16[this->getDatalen()]();
```
gmarkall: What does this change do?

stuartarchibald: Its like `calloc()`, guess it's largely a pointless habit, I can remove if you like.

gmarkall:
This looks great, and looks like it makes things a lot easier/more sensible type hierarchy. I have a few questions (besides the one on the diff):
- The destructors for `OGMatrix` and `OGDiagonalMatrix` are virtual and don't do anything. Can they be removed?
- The constructors for the `OGMatrix`, `OGDiagonalMatrix` and `OGSparseMatrix` all have a variant that sets the data access. It looks like the data access is a property of any `OGArray` and it it could be set once in the `OGArray` constructor, rather than repeated for each subtype. Perhaps the `OGArray` constructor could take the data access, then each subclass constructor could call it in its initialiser list? Since this would be tricky with a separate constructor in each of the subclasses, the data access could be a parameter with a default value in a single constructor for each of the subtypes.
- At the moment all our tests for exceptions use `ASSERT_ANY_THROW`. It would be good to change this to check that we're really throwing the specific exception we expect to throw, so that we can have a little more confidence that the exception being thrown is in the place we think it is thrown. Although it's not too complicated now with just a couple of exceptions, if we add any more in future this becomes more difficult to ensure. If you agree that this would be a good idea, then I'll add a ticket to modify the other `ASSERT_ANY_THROW`s that are not part of this PR.

src/librdag/terminal.cc
stuartarchibald: Patched in a376802.
BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/715](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/715)

stuartarchibald:
Thanks for reviewing. In response.
- Yeah, they can be removed. virtualness will be inherited. Well spotted.
- I agree with the idea of a ctor with a default access parameter, _but_ one could argue the same for the data pointer and {rows,cols}. As noted over IRC we could spend all day talking about this and have mutually agreed to leave it as is and use it more and see if some defining use case comes up that tips the balance one way or the other
- Yes. Will fix. [MAT-343](http://jira.opengamma.com/browse/MAT-343) is the ticket for patching the rest.

stuartarchibald:
Patch 1ada999 addresses using default `DATA_ACCESS` parameter in ctors and removes virtual dtors that are inherited as default.
Patch 04d3e26 swaps `ASSERT_ANY_THROW` for `ASSERT_THROW` with the thrown item specified.

New BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/717](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/717)

gmarkall:
Looks great! Merging.

PR #71, opened 2014-03-18
Selectresult execution implementation

gmarkall:
This PR implements all the mechanisms required for a SELECTRESULT to materialise its result back to Java.

Changes to achieve this consist of:
- Adding a "dummy" SVD node. This node initialises itself with its results, which are `U = OGRealScalar(1.0), S = OGRealScalar(2.0), V = OGRealScalar(3.0)`. This is in lieu of an implementation that calls LAPACK for the SVD computation, and is used to demonstrate/test the SELECTRESULT node.
- A custom runner for SVD nodes is added, which does absolutely nothing, leaving the registers unchanged.
- Changing the Java side SELECTRESULT node to match the C++ size - namely, instead of a special member called `_entry` to store the selected register index, it uses an `OGIntegerScalar`.
- Fixing up the construction of a `JOGIntegerScalar` and associated tests. The JVM Manager has additional members added, which are required to access the `getValue()` function of `OGIntegerScalar`.
- Modification of the generators to generate a specialised `SELECTRESULT` runner and dispatch. It is an exceptional case because its second argument is always an integer scalar. The present implementation looks like a deviation from the rest of the logic in the code generator, but I think we need a few more custom nodes before deciding how they should be handled in a more generic way.
- Addition of `getU()`, `getS()` and `getV()` methods to the Java `SVD` class, which return an appropriate SELECTRESULT, and a test that checks that each of these materialises to the correct result (at the present time, scalar values of 1.0, 2.0, or 3.0 due to the SVD dummyness).

Minor changes:
- Unknown node exceptions now report the node ID that it was not expecting in `createExpr`.
- Occasionally the plus materialise test was smashing the stack on my local machine, so the size of the tree is reduced.
- Removal of the `customrunners.hh` header, which was duplicating generated headers anyway.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/719
Coverage: Java 81% (unchanged), jshim 63.9% (unchanged), rdag 92.4% (unchanged), rdag/runners 69.2% (new, not 100% due to the dummy SVD).


include/expressionbase.hh
```diff
@@ -42,9 +42,9 @@ class OGExpr: public OGNumeric
     virtual const RegContainer * getRegs() const;
   protected:
     OGExpr(ArgContainer* args);
+    RegContainer * _regs;
```
stuartarchibald: Why has this moved from being private? If it is for access, for consistency it probably ought to have a getter instead?

gmarkall: It was so that the SVD could initialise itself with some dummy values in its registers. I didn't really want to add a non-const getter since I thought that could be abused, so this seemed like the safest option without making larger changes.

src/generator/runners.py
```diff
@@ -89,6 +90,56 @@ def complex_matrix_runner_function(self):
...
+        return binary_runner_function % d
+
+    @property
+    def real_matrix_runner_function(self):
+        implementation = self.real_matrix_implementation
+        d = { 'implementation': implementation,
+              'nodename': self.typename,
+              'arg0type': 'OGRealMatrix',
+              'arg1type': 'OGIntegerScalar',
+              'returntype': 'OGRealMatrix' }
```
stuartarchibald: I don't think the return type can't be fixed as a concrete implementation, in fact, I think this is the case more generally too. So for example, SVD would return a diagonal matrix for "S". This also follows for types where the domain may change depending on the numerical domain of execution, for example `atanh()` for real domain inputs `(-1,1)` will return real domain answers elsewhere it'll be complex.

src/java/src/main/java/com/opengamma/longdog/datacontainers/lazy/OGExpr.java
```diff
@@ -45,4 +45,8 @@ public String toString() {
     return this.getClass().getName();
   }
 
+  public OGNumeric getArg(int index) {
+    return _args[index];
```
stuartarchibald, line 54: index ought to be checked for OOB

src/java/src/main/java/com/opengamma/longdog/datacontainers/scalar/OGIntegerScalar.java
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright (C) 2013 - present by OpenGamma Inc. and the OpenGamma group of companies
```
stuartarchibald: year is wrong

src/generator/runners.py
gmarkall: In this case, the return type should presumably be an OGNumeric or OGTerminal, since we could be selecting from absolutely anything in the registers?

src/java/src/test/java/com/opengamma/longdog/testnodes/TestSVDMaterialise.java
```diff
@@ -0,0 +1,80 @@
+/**
+ * Copyright (C) 2013 - present by OpenGamma Inc. and the OpenGamma group of companies
```
stuartarchibald: year is wrong

src/java/src/main/java/com/opengamma/longdog/datacontainers/lazy/OGExpr.java
gmarkall, line 54: and Throw an IndexOutOfBoundsException if it is?

src/generator/runners.py
stuartarchibald: Well I think at the minute everything returns a valid OGTerminal so it's safe to switch to that.

src/java/src/main/java/com/opengamma/longdog/datacontainers/lazy/OGExpr.java
stuartarchibald, line 54: one of `MathsExceptionIllegalArgument.class`

src/java/src/main/java/com/opengamma/longdog/datacontainers/scalar/OGIntegerScalar.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+  @Override
+  public ExprTypeEnum getType() {
+    return s_type;
+  }
+
+  /**
+   * FIXME: getData should not always return double, probably.
```
stuartarchibald: should be a comment not a javadoc

gmarkall: I think I should remove this comment entirely, since it no longer matters now OGIntegerScalar has a getValue() method.

stuartarchibald:
This on the whole this seems sound. There's a few minor bits and bobs as noted above. 
On the larger minor changes front my only observation is:
- Could the svd runner impl wire in the registers opposed to mangling the OGExpr class and hacking SVD in. Whilst it is clearly correct and works, the LAPACK wiring will have to be plumbed into the runner so adding the dummy vars in there I feel would cause minimal impact.

Am happy to merge on patching those small things.

gmarkall:
I've made the following changes:
- `OGTerminal::_regs` is private again
- The `SVD` node has a similar ctor to other nodes now.
- The SVD runner inserts the dummy results (1.0, 2.0, 3.0) into the registers.
- `SelectResultRunner` now always returns an OGTerminal, so any register contents should be safe to have.
- if `getArg` is called with an index out of bounds, a `MathsExceptionIllegalArgument` is thrown, and this behaviour is tested. There is no testing of nodes themselves, only materialise tests, so I have added this to these tests. In future when we revisit the Java side, the node and materialise tests should be split.
- Corrected dates in copyright notices.
- Removed that erroneous javadoc.

New buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/721
Test coverage is as mentionefd above, except that librdag/runners is up to 71.4% due to moving the SVD dummy result generation there.

gmarkall:
I've now:
- Corrected the computation of `lastArg`
- Tightened the index used in the `getArg` exception test so that it would actually spot the off-by-1 that I've just fixed.
- Changed `getV()` to `getVT()` in the SVD node.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/723

stuartarchibald:
Ok, think this is good. Many thanks for working on the patches so quickly. Merging.

PR #72, opened 2014-03-19
Generate expression enum for Java

gmarkall:
This PR generates the Java expression enum, based on a template that has the non-generated nodes hardcoded in it. This is proposed for merge at this time because the majority of hard-to-manage-manually changes will happen with the generated nodes, where the C++ side enum will change automatically, but the Java doesn't. For the non-generated nodes, the C++ side still requires a manual edit, so the need to edit the Java side in these cases is much less surprising (and much easier to work out what the change is).

This PR consists of the following changes:
- Addition of the Java enum template, and code to generate it in the generator.
- Addition of an `exprenum_java` makefile target for regenerating the Java enum file. This should be run, and the result committed, whenever a change is made to the list of nodes in `generator.py`.
- Swapping the use of GLOB for specification of individual files to build for the Java code. This was originally intended to allow generation of the Java enum at build-time only (with no version committed in the repo) but since it is very hard to work with generated code in Java IDEs, this change presently serves no additional purpose for this PR. However, it is good CMake practice, and allows it to track changes in the list of files to be built, so it has been left in as it is safely an improvement to the build system.
- Renaming the Java enum to ExprEnum, and making the enum names the same as on the C++ side, for consistency, and to eliminate the need for two cases in the generator.
- Addition of a note on the generation in the README.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/734
Coverage: Java 81%, rdag 92.4%, rdag/runners 71.4%, jshim 63.9% (all unchanged).


src/generator/enumtemplates.py
```diff
@@ -1,3 +1,10 @@
+#
+# Copyright (C) 2013 - present by OpenGamma Inc. and the OpenGamma group of companies
+#
+# Please see distribution for license.
+#
+
+exprenum_java = """\
 /**
  * Copyright (C) 2013 - present by OpenGamma Inc. and the OpenGamma group of companies
```
stuartarchibald: year needs bumping

Also perhaps a note to say this is auto-generated might be a good idea?

gmarkall: Yes, it is a good idea

stuartarchibald:
This is good and will save a lot of already present confusion and difficulty in mapping the RTTI across languages.

gmarkall:
I've fixed up the copyright year and comment. New buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/738

stuartarchibald:
Thanks for the additional patch. Happy to merge.

PR #73, opened 2014-03-20
Mat 322. Add mtimes() node

stuartarchibald:
This ticket was to add in the `mtimes()`, the PR does this but in the process deals with a load of issues that makes writing templated node implementations easier. A summary of the work:
- The `mtimes()` node is added for both C++ and java.
- The `mtimes()` node is tested for both C++ and java for a selection of common use cases.
- LAPACK calls can now be made via type aware templates such that nodes that rely on their presence can be written in templated form.
- The `norm2()` node impl gets moved into the nodes directory along with the `mtimes()` node impl.
- Code for checking binary nodes via a `TEST_P` was added along with some minor bug fixes in the unary test code.
- The terminal types were adjusted so that templated variants of types could be used more easily.

NOTE: This was all rebased on some heavy upstream churn so the history isn't brilliant!

BB Multi-Pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/759](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/759)
**Coverage:**
Native:
src/jshim 64.1% (up slightly)
src/librdag 92.2% (down slightly)
src/librdag/runners     94.7% (missing bit down to lapack INFO != 0 type problems).
src/librdag/test/nodes 92.2% (missing bit down to SVD not impld).
Java: 82% (up slightly)


src/java/src/main/java/com/opengamma/longdog/helpers/FuzzyEquals.java
```diff
@@ -184,6 +184,37 @@ public static boolean ArrayFuzzyEquals(double[] arr1, double[] arr2)
   }
 
   /**
+   * Checks if two double precision floating point array of arrays are approximately "equal"
+   * Equal means the arrays have values the are considered fuzzy equals appearing in the same order and the arrays the the same dimension.
+   * Default values are used for tolerances
+   * @param arr1 the first value
+   * @param arr2 the second value
+   * of the relative magnitude of the numbers. i.e. invariant of the magnitude of the numbers what is the maximum level of magnitude difference acceptable.
```
gmarkall: I think some lines got interchanged here?

stuartarchibald: ah yes, well spotted, the builders clearly aren't sentient yet and didn't spot that! Will patch.

src/java/src/test/java/com/opengamma/longdog/testnodes/TestMtimesMaterialise.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  @Test
+  public void Real_MatrixMtimesMatrixTest() {
+    double[][] expected = new double[][] { { 14., 32., 50., 68. }, { 32., 77., 122., 167. }, { 50., 122., 194., 266. }, { 68., 167., 266., 365. } };
+    double[][] answer = Materialisers.toDoubleArrayOfArrays(new MTIMES(R_A, R_C));
+    assertTrue(FuzzyEquals.ArrayFuzzyEquals(expected, answer));
+  }
+
+  @Test(expectedExceptions = Exception.class, enabled = false)
```
gmarkall: Could we check for MathsExceptionNativeComputation.class so we are more likely to notice if something else is going wrong here? (e.g. so this fails if there's a convert_error thrown in native)

stuartarchibald: Sure. I fixed the test below but missed this one. I'll also enable it at it currently doesn't run. I think this was written prior to the exception handler and was turned off pending rebase but was never switched back on. Will patch shortly.

gmarkall: Ah, I just noticed this test is disabled - does it not work?

stuartarchibald: Fairly sure it does work, but is off. It is the same as the test [here](https://github.com/OpenGamma/nyqwk2/blob/MAT-322-clean/src/java/src/test/java/com/opengamma/longdog/testnodes/TestMtimesMaterialise.java#L117) but with `typename T = double` when it hits the C++.

stuartarchibald:
8c23c1f addresses https://github.com/OpenGamma/nyqwk2/pull/73#discussion_r10802686 and https://github.com/OpenGamma/nyqwk2/pull/73#discussion_r10803213.
New BB Multi-pass: [http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/760](http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/760)
Disabled test now running: [http://devmaths-lx-1:8011/testng-output/nyqwk2-testing-lnx-289.html#m331](http://devmaths-lx-1:8011/testng-output/nyqwk2-testing-lnx-289.html#m331)

gmarkall:
This looks really good, and although the diff is big, it was nice to read. The only two (minor) things I was wondering about are:
- In the Java mtimes materialisation test, there are tests for V^T*M and Mv for complex matrices but not for real matrices. Also in the C++ tests, there are tests for 1x1 \* 1x1 matrices (i.e. scalar \* scalar in a matrix data structure). Should we have tests for these cases on the Java side as well?
- Calling xgemv when the second arg of mtimes is n x 1 dimensions is a nice optimisation. Is there a similar function that we could use in the case of multiplying a 1_n x n_p pair of matrices (i.e. v^T*M)?

stuartarchibald:
Thanks, and thanks for reviewing. In response:
- I have to admit some of these tests were shipped in from previous iterations of this software ;-) I think the issue here is more of what should we test. Because the graph execution algorithms are "smart" they are far too good at making things "just work" which means there's an insane scope of things that could be tested. Indeed if you take {scalar, vector, vector^T, matrix}, {real, complex} and {4 dominant data structures}, the combinations become prohibitively large to code by hand. In future cases, we should probably have a good long think about automating this. The ticket for discussion is [MAT-347](http://jira.opengamma.com/browse/MAT-347) I'd imagine it will be a long one!
- Errr, yes, come to think of it, I think it's just dgemv() but with `TRANS='T'` not quite sure why I didn't do that! It's now a ticket [MAT-346](http://jira.opengamma.com/browse/MAT-346).

gmarkall:
Addressing both those things in future rather than trying to shove them in this PR sounds good. Thanks for this branch! Merging.

PR #77, opened 2014-03-25
Implement infix runners for matrices

gmarkall:
This enables addition of real and complex matrices to work. Some testing is added, to the extent that the addition operation for scalars is tested. More test cases could be added, but given the requirement to overhaul the java side and testing in general this seemed like work that might be duplicated.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/781
Coverage:
Java: 85% (slight rise)
build/src/librdag: 0.8% (unchanged)
build/src/jshim: 13.3% (slight drop since more code is generated)
src/librdag: 92.2% (unchanged)
src/librdag/runners: 94.7% (unchanged)
src/jshim: 64.1% (unchanged)

### DIFF
--- a/src/generator/runners.py
+++ b/src/generator/runners.py
@@ -1,5 +1,6 @@
 from runnertemplates import runners_header, runners_cc, binary_runner_class_definition, \
                             binary_runner_function, infix_scalar_runner_implementation, \
+                            infix_matrix_runner_implementation,                         \
                             unary_runner_class_definition, unary_runner_function,       \
                             prefix_scalar_runner_implementation, \
                             prefix_matrix_runner_implementation, \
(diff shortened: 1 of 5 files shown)


gmarkall:
The logic to support operations on matrix/scalar, scalar/matrix, (1x1 matrix)/matrix, and matrix/(1x1 matrix) was actually not present to has been added.

A test of each of these four cases for both the real and complex cases has been added. Another test to check for an exception when dimensions mismatch has also been added, since the checking of dimensions is now more permissive to account for 1x1 matrices/scalars. 

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/784

stuartarchibald:
NOTE: The request for the additional implementations mentioned above was made over IRC. Thanks for writing them.

stuartarchibald:
This code looks good. Am happy to merge.

PR #78, opened 2014-03-26
[MAT-337] Catch standard exceptions

gmarkall:
This adds catching of standard exceptions to JShim. Since we only
call standard library functions from with JShim and RDAG, we now
catch all the exceptions that can be thrown.

Because standard exceptions do not carry backtrace information, we
can only translate the exception message into a Java exception and
throw it - it is not possible to provide additional information.

The Java-side `MathsExceptionNativeUnspecified` is used for the
standard exceptions, since we do not know which exception will
have been thrown, and it was not one that we specified.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/787
Coverage:
Java: 82% (unchanged)
build/src/librdag: 13.7% (unchanged)
src/jshim: 62.5% (slightly down due to additional code in untested functions)
src/librdag: 92.2% (unchanged)
src/librdag/runners: 94.7 (unchanged)

### DIFF
--- a/include/jvmmanager.hh
+++ b/include/jvmmanager.hh
@@ -54,6 +54,7 @@ class JVMManager {
     DLLEXPORT_C static jclass getOGComplexSparseMatrixClazz();
     DLLEXPORT_C static jclass getMathsExceptionNativeConversionClazz();
     DLLEXPORT_C static jclass getMathsExceptionNativeComputationClazz();
+    DLLEXPORT_C static jclass getMathsExceptionNativeUnspecifiedClazz();
     DLLEXPORT_C static jmethodID getOGRealScalarClazz_init();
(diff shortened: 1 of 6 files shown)


gmarkall:
Buildbot pass after rebase: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/790

Coverage is unchanged from master, with the exception of jshim, which as mentioned above is down to 62.5% due to the extra code in untested functions.

gmarkall:
Buildbot pass after subsequent rebase: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/792

stuartarchibald:
I'm happy with this. Merging.

PR #79, opened 2014-03-27
[MAT-275] ExecutionList construction is procedural

gmarkall:
This replaces the recursive algorithm with a procedural one that
should work with any size tree.

The algorithm performs a depth-first traversal of the tree, pushing nodes on to
the execution list during the traversal. This is slightly more complicated than
a DFT of a binary tree, since each expression has a variable number of
children. In order to keep track of which child we got to when we revisit an
expression node, we keep a stack of argument positions as well as a stack of
nodes that we are visiting. 

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/789

Coverage:

Java: 85% (unchanged)
build/src/librdag: 12.8% (unchanged)
src/jshim: 64.1% (unchanged)
src/librdag: 92% (slight drop since the ExecutionList code is fully-tested, and is now shorter)
src/librdag/runners: 94.7% (unchanged)

### DIFF
--- a/src/librdag/execution.cc
+++ b/src/librdag/execution.cc
@@ -4,88 +4,99 @@
  * Please see distribution for license.
  */
 
+#include <stack>
 #include "expression.hh"
 #include "terminal.hh"
 #include "execution.hh"
 
 namespace librdag {
 
-class Lineariser: public Visitor
-{
-  public:
-    _ExpressionList* linearise(const OGNumeric* tree)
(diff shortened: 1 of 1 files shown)


stuartarchibald:
This is extremely well written in terms of readability and logic. Indeed the simplicity makes hitting all branches considerably easier. Happy to merge.

PR #82, opened 2014-03-28
Exception fixes.

gmarkall:
This PR is mainly about MAT-359, which fixes the issue where exceptions could be thrown in destructors (which results in undefined behaviour). However, it also contains a MAT-336 change.

MAT-359: Exceptions are prevented from leaving destructors by catching them. The code called in the destructors only throws `convert_error` instances. When an exception is caught, we print a message to standard error, because there is little else we can do. It would be more ideal if we could log this as a message to get picked up elsewhere, but we have no facility for doing this at the moment.

MAT-336: In jshim, visitors are stack-allocated instead of heap-allocated, so we do not need to worry about freeing them ourselves anymore.

I also spotted that we were missing an exception check after a CallIntMethod call.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/802
Coverage:
Java: 85% (unchanged)
build/src/librdag: 13.5% (unchanged)
src/jshim: 61.4% (down slightly since there's more code in the untested jshim parts)
src/librdag: 91.2% (unchanged)
src/librdag/runners: 100% (unchanged)

### DIFF
--- a/src/jshim/jshim.cc
+++ b/src/jshim/jshim.cc
@@ -252,11 +252,9 @@ JNIEXPORT jobjectArray JNICALL Java_com_opengamma_maths_materialisers_Materialis
     const librdag::OGTerminal* answer = entrypt(chain);
     DEBUG_PRINT("Returning from entrypt function\n");
 
-    DispatchToReal16ArrayOfArrays *visitor = new DispatchToReal16ArrayOfArrays();
-    answer->accept(*visitor);
-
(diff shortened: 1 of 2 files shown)


stuartarchibald:
I'm happy with this. Added [MAT-364](http://jira.opengamma.com/browse/MAT-364) to look into logging. Merging.

stuartarchibald:
[MAT-364](http://jira.opengamma.com/browse/MAT-364)  is a duplicate of [MAT-360](http://jira.opengamma.com/browse/MAT-360), marking as dupe and closing.

PR #85, opened 2014-04-03
[MAT-246] ArgContainers constructed by expressions

gmarkall:
It is no longer necessary to construct an ArgContainer outside
an expression constructor. Instead, expressions take their actual
arguments and build the ArgContainer themselves, in order to
simplify memory management.

The use of a vector<const OGNumeric*>\* for passing the results of
generateArgs() is a temporary measure, and will be removed when
the translation is converted to a procedural implementation.

Some missing exception checks are also added as they were spotted
during modification of the JNI interface.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/817
Coverage: 
Java: 91% 
build/src/jshim: 0.6% (unchanged)
build/src/librdag: 13.5% (unchanged)
src/jshim: 61.3% (down 0.1% due to a couple of extra lines in untested regions)
src/librdag: 90.9% (down 0.6% due to slightly shrinking the portion of code that was tested)
src/librdag/runners: 100% (unchanged)

### DIFF
--- a/src/generator/exprtemplates.py
+++ b/src/generator/exprtemplates.py
@@ -51,13 +51,19 @@ expr_class = """\
 class %(classname)s: public %(parentclass)s
 {
   public:
-    %(classname)s(ArgContainer *args);
+%(constructor)s
     virtual OGNumeric* copy() const override;
     virtual const %(classname)s* as%(classname)s() const override;
     virtual void debug_print() const override;
     virtual ExprType_t getType() const override;
 };
 """
+
+unary_constructor = """\
+    %(classname)s(const OGNumeric* arg);"""
+
+binary_constructor = """\
+    %(classname)s(const OGNumeric* arg0, const OGNumeric* arg1);"""
 
 # Expressions cc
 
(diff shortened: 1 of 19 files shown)


src/generator/exprtemplates.py
```diff
@@ -119,6 +125,18 @@ class %(classname)s: public %(parentclass)s
...
 """
 
+unary_ctor_method = """\
+%(classname)s::%(classname)s(const OGNumeric* arg): OGUnaryExpr{arg} {}"""
+
+binary_ctor_method = """\
+%(classname)s::%(classname)s(const OGNumeric* arg1, const OGNumeric* arg2): OGBinaryExpr{arg1, arg2} {}"""
+
+unary_copy_method = """\
+  return new %(classname)s((*_args)[0]->copy());"""
```
stuartarchibald, line 135: I think at some point replacing `(*_args)[int]` with a `getArg(int foo)` would be nice :) Doesn't need to go in now.

gmarkall, line 135: Yes, it would - maybe as `getArg(size_t arg)`.

stuartarchibald:
I think this is a considerable improvement. Although the diff is large, most of the changes are very similar and the reduction in code size and complexity brought in by this change is considerable. Happy to merge as WIP towards exception safety in tree translation.

PR #86, opened 2014-04-04
[MAT-372] PtrVector: ownership a runtime property

gmarkall:
OwningPtrVector and NonOwningPtrVector are collapsed into the
single PtrVector type. A PtrVector always begins life as a non-
owning entity. It can then be made to own its elements later on.

This is because it is almost never desirable to own elements
immediately, because some work may have to be done to populate it,
which could throw an exception.

Buildbot pass pending - waiting on suitable Mac to build on.
Change-Id: I4a5f82c6af193dcd54d8ba41230b68eafffb14f7

### DIFF
--- a/src/librdag/containers.cc
+++ b/src/librdag/containers.cc
@@ -15,15 +15,18 @@ namespace librdag {
  */
 
 template<typename cTp>
-PtrVector<cTp>::PtrVector()
-{
-  _vector = new vector<cTp>();
-}
+PtrVector<cTp>::PtrVector(): _vector{}, _owning{false} {}
 
 template<typename cTp>
 PtrVector<cTp>::~PtrVector()
 {
-  delete _vector;
+  if (_owning)
+  {
+    for (auto it = this->begin(); it != this->end(); ++it)
+    {
+      delete *it;
+    }
+  }
 }
 
 template<typename cTp>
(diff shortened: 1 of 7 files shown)


gmarkall:
This probably doesn't want merging yet, even if you get the chance to test it next week. I pushed some additional dependent commits (that should be in separate PRs) on top of the original one that was for review. Will make separate PRs next week.

stuartarchibald:
Acknowledged. Shan't review until bumped. Hopefully the mac can be repaired. :-\

gmarkall:
The Mac is working now and is able to get through a build without its wheels coming off. This branch now also includes the small fix required for imported targets on CMake 2.8.12.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/845
Coverage:
Java: 91%
build/src/jshim: 0.6%
build/src/librdag: 13.5%
src/jshim: 61.3%
src/librdag: 90.7%
src/librdag/runners: 100%

All unchanged except src/librdag, which is down slightly because there is now only one container implementation, rather than two.

src/librdag/containers.cc
```diff
@@ -31,126 +34,122 @@ void
...
  * data.
  */
 
 template<typename cTp, bool Q = is_fundamental<typename remove_pointer<cTp>::type>::value >
-struct owningptrvector_copy_impl;
+struct ptrvector_copy_impl;
 
 template<typename cTp>
-struct owningptrvector_copy_impl<cTp, false>
+struct ptrvector_copy_impl<cTp, false>
```
stuartarchibald, line 107: Please could a unit test be added to cover this template specialisation. The logic seems fine so would be largely for regression purposes. Many thanks.

stuartarchibald:
Subject to the test noted above being written, I think this code is good and will be happy merge.

gmarkall:
I've added a relevant test. Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/847

Librdag coverage is now up to 91.4%, and containers.cc coverage is 100%.

stuartarchibald:
Test looks good. Thanks for doing that. Happy to merge.

PR #87, opened 2014-04-10
Make _args and _regs non-pointers

gmarkall:
This simplifies memory management, as part of the process of avoiding leaks when exceptions are thrown. Since pointers to these members were previously returned, we now return references to them, this also simplifies syntax in a lot of places, reducing the need for explicit dereferences.

Rationale: 
- There is a 1:1 relationship between ArgContainers and Exprs, so we don't need to use pointers to hold ArgContainers.
- There is also a 1:1 relationship between RegContainers and Exprs. This may change in future if we adopt some more advanced allocation scheme - however for now, it is better to have a simpler implementation rather than a future-proof one, especially given the difficulty of testing jshim code.

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/848
Coverage:
Java: 91%
build/src/jshim: 0.6%
build/src/librdag: 13.5%
src/jshim: 61.3%
src/librdag: 91.3%
src/librdag/runners: 100%

All unchanged except librdag, which is down 0.1% because we don't have code for creating and deleting `_args` and `_regs` anymore

### DIFF
--- a/include/expressionbase.hh
+++ b/include/expressionbase.hh
@@ -35,16 +35,16 @@ class OGExpr: public OGNumeric
   public:
     virtual ~OGExpr();
     const ArgContainer& getArgs() const;
+    RegContainer& getRegs() const;
     size_t getNArgs() const;
     virtual const OGExpr* asOGExpr() const override;
     virtual void debug_print() const override;
     virtual void accept(Visitor &v) const override;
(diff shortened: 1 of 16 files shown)


stuartarchibald:
This looks good. Thank goodness we now have `mutable` at our disposal. Merging.

PR #88, opened 2014-04-10
[MAT-336] Procedural expression convertor

gmarkall:
Cleaning up when an exception occurs is not yet implemented. However, this is written such that cleaning up is possible. The traversal algorithm is similar to that used by the execution list linearisation, but the implementation is different enough due to all the JNI things that need to be done that I thought it would be counterproductive to try and use a single implementation for both.

The conversion happens in two stages - in the first stage, all the information required to construct the new expression is gathered, and is stored in a `vector` of `jnode` objects. In the second stage, this vector is iterated along to construct the RDAG tree. 

Buildbot pass: http://devmaths-lx-1:8011/builders/nyqwk2-testing/builds/851

Coverage:

Java: 91%
build/src/jshim: 0.8%
build/src/librdag: 13.5%
src/jshim: 57.7%
src/librdag: 91.3%
src/librdag/runners: 100%

All unchanged except for src/jshim, which has dropped because the expression translation is not tested from the C++ side, and the procedural translation in two stages is much longer than the original recursive implementation.

### DIFF
--- a/include/exprfactory.hh
+++ b/include/exprfactory.hh
@@ -13,8 +13,7 @@ using librdag::OGNumeric;
 
 namespace convert {
 
-vector<const OGNumeric*>* generateArgs(jobject obj);
 OGNumeric* createExpression(jobject obj);
-OGNumeric* createExprWithID(jlong ID, jobject obj);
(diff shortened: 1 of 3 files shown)


stuartarchibald:
I am happy with this on the whole. Clearly full testing is pending the `JVMManager` class rewrite. I particularly appreciated the running commentary in the expression tree translation algorithm. Merging.

PR #89, opened 2014-04-14
Mat 368 cont.

stuartarchibald:
This PR:
- Adds in a number of functions from LAPACK/BLAS via templates. The purpose of doing this is so that we can write methods in terms of templates that will use LAPACK/BLAS for the lifting without worrying about
  - what the data type is
  - F77 calling conventions 
  - computing the workspace requirements/memory management
- Also included are patches to make the LAPACK frames exception safe in terms of memory management.

Coverage:
Java: 91%
build/src/librdag 13.5% (no change)
src/jshim 57.7% (no change)
src/librdag 92.9% (up 1.6%)
src/librdag/runners 96.3% (down 3.6%, exception catches not covered in SVD runner)

### DIFF
--- a/include/lapack.hh
+++ b/include/lapack.hh
@@ -155,7 +155,7 @@ template<typename T> void xgemm(char * TRANSA, char * TRANSB, int * M, int * N, 
 /**
  * xnrm2 generalised 2-norm implementation
  * @param N as BLAS dnrm2 N
- * @param A data type specific with intent as BLAS dnrm2 A
+ * @param X data type specific with intent as BLAS dnrm2 X
  * @param INCX as BLAS dnrm2 INCX
  * @return the 2-norm of X
  */
@@ -172,7 +172,6 @@ template<typename T> real16 xnrm2(int * N, T * X, int * INCX);
  * @param A data type specific with intent as LAPACK dgesvd A
  * @param LDA as LAPACK dgesvd LDA
  * @param S as LAPACK dgesvd S
- * @param T as LAPACK dgesvd T
  * @param U as LAPACK dgesvd U
  * @param LDU as LAPACK dgesvd LDU
  * @param VT as LAPACK dgesvd VT
--- a/include/lapack.hh
+++ b/include/lapack.hh
@@ -37,6 +37,33 @@ extern "C" real16 F77FUNC(dznrm2)(int * N, complex16 * X, int * INCX);
 extern "C" void F77FUNC(dgesvd)(char * JOBU, char * JOBVT, int * M, int * N, real16 * A, int * LDA, real16 * S, real16 * U, int * LDU, real16 * VT, int * LDVT, real16 * WORK, int * LWORK, int * INFO);
 extern "C" void F77FUNC(zgesvd)(char * JOBU, char * JOBVT, int * M, int * N, complex16 * A, int * LDA, real16 * S, complex16 * U, int * LDU, complex16 * VT, int * LDVT, complex16 * WORK, int * LWORK, real16 * RWORK, int * INFO);
 
+// Reciprocal condition number estimates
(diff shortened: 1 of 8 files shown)


include/lapack.hh
```diff
@@ -95,32 +109,169 @@ extern complex16 * czero;
...
+ * @param LDC as BLAS dgemm LDC
+
  */
 template<typename T> void xgemm(char * TRANSA, char * TRANSB, int * M, int * N, int * K, T * ALPHA, T * A, int * LDA, T * B, int * LDB, T * BETA, T * C, int * LDC );
 
 /**
- * xnrm2 general norm2 implementation
+ * xnrm2 generalised 2-norm implementation
+ * @param N as BLAS dnrm2 N
+ * @param A data type specific with intent as BLAS dnrm2 A
```
gmarkall: There is no parameter`A`?

stuartarchibald: oops. That should be `X`, will patch.

include/lapack.hh
```diff
@@ -95,32 +109,169 @@ extern complex16 * czero;
...
- * xgesvd general svd implementation that takes care of workspaces
+ * xgesvd is a generalised svd implementation that takes care of workspaces
+ * @param JOBU as LAPACK dgesvd JOBU
+ * @param JOBVT as LAPACK dgesvd JOBVT
+ * @param M as LAPACK dgesvd M
+ * @param N as LAPACK dgesvd N
+ * @param A data type specific with intent as LAPACK dgesvd A
+ * @param LDA as LAPACK dgesvd LDA
+ * @param S as LAPACK dgesvd S
+ * @param T as LAPACK dgesvd T
```
gmarkall: There is no parameter `T`

stuartarchibald: I'm going to take a look at these. I was clearly over enthusiastic on the `sed` front.

src/librdag/runners/svdrunner.cc
```diff
@@ -39,12 +39,27 @@ template<typename T> void svd_dense_runner(RegContainer& reg, const OGMatrix<T>*
...
-  lapack::xgesvd(lapack::A, lapack::A, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
+  try
+  {
+    lapack::xgesvd(lapack::A, lapack::A, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
+  }
+  catch (exception& e)
+  {
+    if(info < 0) // illegal arg
+    {
+      delete[] A;
```
gmarkall, line 50: Do `U`, `S` and `VT` also need deleting here to prevent leaks, since they don't get passed to the constructor of an owning terminal, and their pointers will subsequently be lost?

gmarkall:
This looks great. Apart from the comments inline, I have one other question - why does every lapack routine individually call `set_xerbla_death_switch()` - could it not just be called once, in jinitilialise?

src/librdag/runners/svdrunner.cc
stuartarchibald, line 50: Yes. Though I'm considering the merits of even having this branch. I think it's dead as the `OGMatrix<T>` ctors refuse negative args and the unit tests for this code pass, which suggests that all the parameters and plumbing are correct and this branch can't be tripped.

stuartarchibald:
Were it only called once it'd require all unit testing paths to run through `jinitialise` which is why its set everywhere locally as most unit tests rightfully don't know about java.

src/librdag/runners/svdrunner.cc
gmarkall, line 50: I think it's worth leaving in, for the case when something changes that logic or we find there's some oversight in the logic somewhere.

stuartarchibald:
Patches in 0fa9687.

gmarkall:
This looks great - good to have everything covered along with the new Lapack interfaces. Happy to merge.

PR #91, opened 2014-04-16
[MAT-378] Version info in JAR, fix rebuilding

gmarkall:
Rebuilding of the JAR after some a non-java file that is included
in it changes should now be fixed. This is achieved by always
rebuilding a JAR containing native code, by using a custom target
to build the JAR instead of a custom command.

Also, handling of the DEPENDS argument to add_jar() is implemented
- a DEPENDS argument was specified in the main JAR's add_jar, but
  DEPENDS was not actually being handled by add_jar.

A small script to generate a text file containing the SHA1 of the
commit for inclusion into the JAR is added. The correct updating
of this file every time the JAR is built is handled by using the
new DEPENDS argument.

Coverage is unchanged - this is a build system change only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,6 +197,20 @@ endif()
 # Where the generated include files live
 set(BIN_INCLUDE_DIR ${CMAKE_BINARY_DIR}/include)
 
+# Add version info file
+if(WIN32)
+  set(PLATFORM_CODE win)
+else()
+  if(APPLE)
+    set(PLATFORM_CODE osx)
+  else()
+    set(PLATFORM_CODE lnx)
+  endif()
+endif()
+add_custom_target(verinfo
+                  ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/cmake/verinfo.py ${CMAKE_BINARY_DIR}/lib/og-maths-${PLATFORM_CODE}-sha1.txt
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -197,6 +197,10 @@ endif()
 # Where the generated include files live
 set(BIN_INCLUDE_DIR ${CMAKE_BINARY_DIR}/include)
 
+add_custom_target(verinfo
+                  ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/cmake/verinfo.py ${CMAKE_BINARY_DIR}/lib/og-maths-sha1.txt
```
stuartarchibald: Any chance the sha1 text file could be named after platform as well, purely to prevent collisions if the build ends up in a tangle?

cmake/verinfo.py
```diff
@@ -0,0 +1,18 @@
+import sys, subprocess
```
stuartarchibald, line 7: Please add license/copyright header to this.

gmarkall:
I've updated the PR to include the copyright header and name the version file including the platform name. This now includes lnx for Linux, osx for OS X, and win for Windows, e.g. `og-maths-lnx-sha1.txt`.

stuartarchibald:
This looks good, will certainly cut down the coding building cycle time. Happy to merge.

PR #94, opened 2014-04-25
Mat 381 - create pinv() node

stuartarchibald:
The purpose of this PR is to add in the pseudo-inverse operation. It is this operation that is needed in platform for various legacy behaviour replication reasons. It...
- Adds `PINV` for pseudo-inverse, `TRANSPOSE` for transpose, and `CTRANSPOSE` for conjugate transpose. Conjugate transpose is needed in the computation of `PINV` and normal transpose is a trivial simplification so I added it anyway. The `PINV` node is largely written in terms of other nodes :)
- Adds tolerances via overloads to `OGTerminal::mathsequals()` and  `OGTerminal::fuzzyequals()` methods.
- Standardises the `FuzzyEquals` defaults, fixes an associated bug and patches the test.
- Adds the `runtree` function that sets up and runs trees from a given root node.
- Implements missing `toComplex16ArrayOfArrays()` override for `OGRealScalar` type.
- Adds a load of C++ and Java tests of pure checked input/output pairs AND where applicable reconstruction tests.
- Adds code to the `XXX_NODE_TEST_SETUP` macro to print debug statements on comparison failure.

Coverage:
Java: 91% (no change)
build/src/librdag 15.2% (up 1.7%)
src/jshim 57.9% (up 0.2%)
src/librdag 91.5% (down 1.4%, this is due to the required pure methods on the `NODE` impls not being tested, this is scheduled for autogen)
src/librdag/runners 96.7% (up 0.4%)


include/terminal.hh
```diff
@@ -430,6 +456,15 @@ OGTerminal * makeConcreteDenseMatrix(real16 * data, int rows, int cols, DATA_ACC
 template<>
 OGTerminal * makeConcreteDenseMatrix(complex16 * data, int rows, int cols, DATA_ACCESS access);
 
+template<typename T>
+OGTerminal * makeConcreteScalar(T data);
```
gmarkall, line 473: Can you document this method please? I'm not sure what "concrete" means just from the name.

stuartarchibald, line 473: Sure, will patch. I think there's another "makeConcreteFoo" floating about, I'll make sure that's done too.

src/jshim/jdispatch.cc
```diff
@@ -146,7 +146,9 @@ void DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<complex16> const
 }
 void DispatchToComplex16ArrayOfArrays::visit(librdag::OGScalar<real16> const SUPPRESS_UNUSED  *thing)
```
gmarkall: thing is not unused

stuartarchibald: Well spotted, will patch.

src/librdag/terminal.cc
```diff
@@ -1517,5 +1554,24 @@ OGTerminal * makeConcreteDenseMatrix(complex16 * data, int rows, int cols, DATA_
   return new OGComplexMatrix(data, rows, cols, access);
 }
 
+// Concrete template factory for scalars
+
+template<typename T>
```
gmarkall: I suspect the generic implementation isn't necessary - with only the header file, another piece of code wouldn't be able to instantiate the template for a type other than a real16 or a complex16, giving a compile-time error anyway.

src/librdag/test/check_equals.cc
```diff
@@ -48,7 +48,7 @@ TEST(EqualsTest, SingleValueFuzzyEqualsReal16) {
   ASSERT_TRUE(SingleValueFuzzyEquals(neg0.d,neg0.d));
 
   // same value as it trips the return true on "difference less than abs tol" branch
-  ASSERT_TRUE(SingleValueFuzzyEquals(DBL_EPSILON,2.e0*DBL_EPSILON));
+  ASSERT_TRUE(SingleValueFuzzyEquals(1.1e0*DBL_EPSILON,2.e0*DBL_EPSILON));
```
gmarkall, line 51: Is this loosened because some default changed elsewhere?

src/librdag/test/nodes/check_ctranspose.cc
```diff
@@ -0,0 +1,105 @@
...
+
+class ReconstructCtransposeNodeTest: public ::testing::TestWithParam<const OGTerminal*> {};
+
+TEST_P(ReconstructCtransposeNodeTest, TerminalTypes)
+{
+  const OGTerminal * A = GetParam();
+  OGExpr * ct = new CTRANSPOSE(A);
+  OGExpr * ctctA = new CTRANSPOSE(ct); // use copy else there's two refs to one terminal floating about
+  runtree(ctctA);
+  // this is temporary pending ->mathsequals() getting tolerance options
```
gmarkall: Mathsequals does appear to have tolerance options now - what would/should change?

src/librdag/terminal.cc
stuartarchibald: Fair point. Will remove.

gmarkall:
This looks good - I only have the few inline comments, and I'm happy to merge this once they're resolved.

src/librdag/test/check_equals.cc
stuartarchibald, line 51: The code read as:

```
bool SingleValueFuzzyEquals(real16 val1, real16 val2, real16 maxabserror = 10*DBL_EPSILON, real16 maxrelerror = 10*DBL_EPSILON);
```

```
bool SingleValueFuzzyEquals(complex16 val1, complex16 val2, real16 maxabserror = DBL_EPSILON, real16 maxrelerror = 10*DBL_EPSILON);
```

Such that the `maxabserror` was different depending on whether you were comparing `real16` or `complex16`. The change to `equals.cc` patches this and this test is adjusted so the diff of `(2.0-1.1)*DBL_EPSILON <  FuzzyEquals_default_maxabserror = std::numeric_limits<real16>::epsilon()` trips the branch.

src/librdag/test/nodes/check_ctranspose.cc
stuartarchibald: I shall remove this comment.

include/terminal.hh
stuartarchibald, line 473: Patched. 526974c

src/jshim/jdispatch.cc
stuartarchibald: Patched 7c0a989 .

src/librdag/terminal.cc
stuartarchibald: Patched c5933e7

src/librdag/test/nodes/check_ctranspose.cc
stuartarchibald: Patched 85bc582

stuartarchibald:
Many thanks for reviewing. I think all comments are now addressed.

gmarkall:
Looks great - merging.

PR #96, opened 2014-05-01
Mat 383 Add LU

stuartarchibald:
This PR adds code which implements the `LU` decomposition. The `LUP` decomposition was offered but rejected. The decomposition uses `LAPACK`'s `{d,z}getrf` function to perform the decomposition. The information from`{d,z}getrf` regarding the pivots is then applied to `L`, such that the `LU` decomposition of matrix `A` reconstructs via `A = LU`. This behaviour was chosen in preference to `P*A = L*U` such that `A = P**T * L * U`, where `P` is a permutation matrix and `**T` indicates matrix transpose.

This PR also:
- Adds the LAPACK function `xswap()`, I thought I was going to use this then didn't. As it is a useful function I've kept it in, tests are included.
- Adds the LAPACK function `xgetrf()` for performing `LU` decompositions.
- Adds to Java the `OGLUResult` class to handle the graph building of an `LU` decomposition result.
- Adds tolerance options to the Java `FuzzyEquals.ArrayFuzzyEquals()` method operating on `double[][]`.
- Adds the `lu()` function to `DOGMA`.
- Adds a note regarding missing logging for detection of singular matrices. The message is correct and written to `STDERR`, the unit tests also cover this branch. Adding in a true check should be relatively easy once the upstream code is written.
- Does some general housekeeping regarding missing test coverage. I think a templated approach to covering `VirtualCopyTest` type tests would be a good idea for the future, see [[MAT-392]](http://jira.opengamma.com/browse/MAT-392).

Coverage:
Java: 91% (no change)
build/src/librdag 15.7% (up 0.5%)
src/jshim 57.9% (no change)
src/librdag 92.1% (up 0.6%)
src/librdag/runners 96.0% (down 0.7%, largely due to LAPACK bad-input exception handling not being triggered).


gmarkall:
This looks great - thanks for bringing up the test coverage with those other bits of functionality as well. A templated approach for those tests sounds like a good idea.

I've reviewed the code as best I can with my limited understanding of the LU factorisation algorithm. It looks sane, but I can't guarantee that I'm able to think about all the edge cases. That said, the testing of the Lapack interface, the Rdag C++ side and the Java interface all looks thorough and solid, so I feel comfortable that the code has the requisite degree of robustness.

Thanks for patching the assignment of `lda` also - I think this PR is good, and ready for a merge.

PR #97, opened 2014-05-02
Add GCC runtime libraries into JARs

gmarkall:
This PR provides only the mechanism required to get the GCC runtime libraries into the Jar on each platform. It does not attempt to fix up the linkage of these libraries - this will be done in a future PR.

The library locations are first retrieved from the CMake argument `GCC_LIB_FOLDER`. If this is not specified, the `GCC_LIB_FOLDER` environment variable is checked. If this is also not present, a default library location is assumed (which varies by platform and matches those used on our buildslaves).

Code for computing the platform code is outlined in both the Python and CMake scripts, since it is now required from multiple locations.

Relevant license information is also added.

### DIFF
--- a/cmake/verinfo.py
+++ b/cmake/verinfo.py
@@ -9,7 +9,7 @@ the current platform. Version info is a yaml file containing
 the revision number, build number, version number, project name,
 timestamp/local time, and list of artifacts."""
 
-import platform, sys, subprocess, time
+import sys, subprocess, time
 from buildutils import platform_code
 from yaml import load, dump
 try:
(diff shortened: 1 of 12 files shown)


cmake/verinfo.py
```diff
@@ -7,9 +7,10 @@
 """This tool creates a version info file for an OG-Maths build on
 the current platform. Version info is a yaml file containing
 the revision number, build number, version number, project name,
-and list of artifacts."""
+timestamp/local time, and list of artifacts."""
```
stuartarchibald, line 10: thanks.

stuartarchibald:
I'm happy with this PR. I believe it covers what was asked of us. Merging.

PR #98, opened 2014-05-06
Fix linkage of included libraries on Linux and OS X

gmarkall:
This is heavily based on the DOGMA1 implementation of linkage-fixing, in that the operations it performs are identical, but are performed as part of the local build rather than requiring them to be implemented by an external entity like the Buildbot.

This now allows the multiplatform JAR to execute "out of the box" on completely clean Linux, OS X, and Windows machines (assuming that a JRE is present, of course), and I have verified this manually on devmaths-mc-2, devmaths-wn-2 and devmaths-lx-cleantest. Automation of this testing is to be performed in the near future.

Coverage is unchanged, as this is only a build system-related change.

The build of deploy that this was tested with is http://devmaths-lx-1:8012/builders/deploy/builds/33 - the resulting blob can be downloaded and run on any supported machine with no "hacks" required.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -39,7 +39,7 @@ Generally recent versions of the requirements are preferred:
 * gfortran (we test with GCC 4.8.2)
 * A C++ compiler supporting C++11 (we test with GCC 4.8.2)
 * A C compiler supporting C99 (we test with GCC 4.8.2)
-* Python 2.7 with argparse and yaml modules (we test with Python 2.7)
+* Python 2.6 or 2.7 with argparse and yaml modules (we test with Python 2.6 on Linux and 2.7 on Windows/OS X)
 * JDK 7 (we test with Oracle JDK 1.7.0_51 and 1.7.0_u21)
 * Maven (we test with 3.0.5)
 
--- a/cmake/modules/UseJava.cmake
+++ b/cmake/modules/UseJava.cmake
@@ -395,7 +395,7 @@ function(add_jar _TARGET_NAME)
       set_platform_code(NATIVE_PLATFORM)
       set(_copyrtl_args)
       if (GCC_LIB_FOLDER)
-        set(_copyrtl_args "-l ${GCC_LIB_FOLDER}")
+        set(_copyrtl_args "-l${GCC_LIB_FOLDER}")
       endif()
       add_custom_target(copyrtl
                         ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/cmake/syslibs.py -o ${_dest}/${NATIVE_PLATFORM} ${_copyrtl_args}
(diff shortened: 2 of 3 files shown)


stuartarchibald:
Although not related to the contents of this PR, whilst checking the resulting blobs, there appears to be an issue about order of writing in the `verinfo.yaml`. The info for each category is written as `[lnx, osx, subprojects*, win]`, whilst alphabetical, is slightly odd to read! Ticket [[MAT-396]](http://jira.opengamma.com/browse/MAT-396).

stuartarchibald:
NOTE: I have just manually checked the jar on linux. `RUNPATH` is as expected, jar executes as expected and all symbols in native libraries appear to be defined. Native linkage in general is `$ORIGIN`, exceptions being `libm`, `libc`, `libpthread`, `ld-linux` and `linux-vdso` as expected.

stuartarchibald:
NOTE: Manual check on another linux VM (`fedora64bit_mathstesting`, which is local to me) also passes.

stuartarchibald:
NOTE: Manual check on windows machine with `%PATH%` unset and then running the java binary from its path to binary directly also passes.

stuartarchibald:
NOTE: Manual check on OSX produces expected result. `otool -L` correctly reports `@loader_path` for library linkage. Libraries depend on `libSystem.B.dylib`, we should discuss this as I rather suspect OSX older than our build system may have versions of this with things missing, fair enough, but we need to make sure upstream `A{P,B}I`'s in newer `libSystem`s are compatible OR make declarations on support.

stuartarchibald:
Bug: Whilst I realise that this isn't in the current change set, it is related as it is the local input to the code which is in the change set. I just tried building this branch end to end locally. I think this is a bug: https://github.com/OpenGamma/OG-Maths/blob/master/cmake/modules/UseJava.cmake#L398

```
set(_copyrtl_args "-l ${GCC_LIB_FOLDER}")
```

the space between `-l` and `${GCC_LIB_FOLDER}` ends up in the arg to `syslibs.py` e.g. `[SPACE]/usr/liib64` which means the libs can't be found as they are all prefixed with a space!

stuartarchibald:
I think the above bug related to this code (assuming my guess is correct) needs fixing.  Perhaps we should have a builder that runs with the `-D` options too? Other than that, I'm happy for this code to go into mainline, good job.

gmarkall:
Thanks for spotting that bug - I hadn't previously spotted it because I was testing syslibs.py on the command line, so I when I ran syslibs.py -l <loc>, this was the equivalent of "syslibs.py" "-l" "<loc>", whereas CMake is not running the command through the shell, so the argument was being sent as "syslibs.py" "-l loc" causing the argument parser to parse in the space.

This should now be fix. I also added a small documentation fix as I noticed the required versions of Python and those that we test with were not listed correctly.

stuartarchibald:
Great. Thanks for fixing that. Merging.

PR #103, opened 2014-05-28
Mat 386 Implement Fuzzer

stuartarchibald:
This PR is for addition of a rough implementation of a fuzzer  cf. those found in penetration testing.
- Adds fuzzer capabilities via the creation of random data and execution of random trees. This is performed by:
  - Reflection on the `DOGMA` API.
  - Creation of random `OGTerminal` types.
  - Glueing together the above into trees which are then executed.
  - Caching parts of the trees in a fixed size FIFO store so that they can become inputs to later fuzzes. This also prevents JVM heap bloat.
- Additional work is needed to create a better rule set about the generated data, most trees fail due to non-commuting data over the operation attempted.
- The fuzzer can be run in a threaded mode and has options regarding the number of refs and data sizes used in the trees as well as the number of threads used.
- Logging capabilities are provided such that trees are written out to logs before execution. This way, if a catastrophic error occurs where no recovery is possible a log is present so the tree can be recreated and debugged.
- A command line tool is also provided. Documentation of this tool needs providing during the documentation sweep. An example invocation is:

```
java -cp ${CLASSPATH}:og-maths-lnx-0.1.0-SNAPSHOT.jar 'com.opengamma.maths.fuzzer.FuzzerMain' -t 60 -c 4  -l -p
```

which runs the fuzzer for 60 seconds over 4 threads with logging and a statistics pretty print enabled.
- An annotation transformer is used to prevent the unit tests from running unless `-DrunFuzzerTests` is specified on the jvm command line.
- The fuzzer actually found a bug whilst I was writing it, [MAT-404](http://jira.opengamma.com/browse/MAT-404). This is now patched.

Coverage:
Java: 79% (down 12%, this is due to the fuzzer tests being disabled as they basically cause false coverage as they execute large numbers of random paths).
Native: no change


src/java/src/main/java/com/opengamma/maths/fuzzer/DataCreator.java
```diff
@@ -0,0 +1,135 @@
...
+    data[0][0] = 1; // set this to prevent entirely empty matrices
+    return data;
+  }
+
+  /**
+   * Next int, will be > 0
+   * @return
+   */
+  private int intGt0() {
+    return _rand.nextInt(_maxsize) + 1;
```
gmarkall, line 93: I think there is the possibility of an integer overflow here - if _rand.nextInt returns the maximum int, this function will return 0?

gmarkall, line 93: (if maxsize is the maximum integer, of course!)

gmarkall, line 93: Actually, perhaps not, since maxsize is excluded from the distribution. Apologies!

src/java/src/main/java/com/opengamma/maths/fuzzer/FuzzerMain.java
```diff
@@ -0,0 +1,185 @@
...
+
+    // provision of logging dir switches on logging
+    if (loggingOn || logDirStr != null) {
+      loggingOn = true;
+      // -l switches on logging, but log dir may not have been provided, create one
+      try {
+        if (logDirStr == null) {
+          Path tmpDir;
+          tmpDir = createTempDirectory("og-maths-logs-");
+          logDir = tmpDir.toFile();
```
gmarkall, line 171: Specifying the log directory results in a NullPointerException - I guess you also want to create logDir when it is not the default?

stuartarchibald, line 171: That's a fair point, writing to a directory is always easier if it exists! Will patch.

stuartarchibald, line 171: Patched in ad1a6d4

gmarkall:
This PR looks good - the fix for the NullPointerException appers to do the job. It's good that the fuzzer already spotted a bug, and this will provide a good foundation for testing future memory management fixes. Happy to merge.

PR #104, opened 2014-05-30
Mat 329/330

stuartarchibald:
This PR adds patches for a couple of issues:
- Removes all caching of the JNI environment pointer. Even though the actual execution using the pointer was safe via the ordering in which it was performed, it was rather confusing. Removing the cache and always obtaining a local JNIEnv pointer regardless of context means the code is much clearer at the cost of a tiny loss in performance.
- Disables tests that rely on the impairing of `AttachCurrentThread()` to break the `getEnv()` mechanism, all `JNIEnv` calls now go through the `getEnv()` function now, including those which are statically initialised. Therefore causing impairment to the `AttachCurrentThread()` function breaks the initialisers before the code to be tested can be run.
- Deletes a local ref that escaped. There's no harm in leaving it but might as well keep things clean if possible.

Coverage:
Java: 79% (no change)
build/src/librdag 16.1%  (no change)
src/jshim 57.6% (down 0.3%, due to method removal)
src/librdag 91.8% (no change)
src/librdag/runners 79% (no change).

### DIFF
--- a/src/jshim/jvmmanager.cc
+++ b/src/jshim/jvmmanager.cc
@@ -73,70 +73,70 @@ JVMManager::initialize(JavaVM *jvm)
 
   // Set up cached pointers
   _jvm = jvm;
-  registerReferences();
+  registerReferences(env);
 }
 
 void
-JVMManager::registerReferences()
+JVMManager::registerReferences(JNIEnv * env)
 {
   //
   // REGISTER CLASS REFERENCES
   //
 
-  registerGlobalClassReference("java/lang/Double", &_DoubleClazz);
-  registerGlobalClassReference("com/opengamma/maths/datacontainers/OGNumeric", &_OGNumericClazz);
-  registerGlobalClassReference("com/opengamma/maths/datacontainers/OGTerminal", &_OGTerminalClazz);
-  registerGlobalClassReference("com/opengamma/maths/datacontainers/matrix/OGArray", &_OGArrayClazz);
-  registerGlobalClassReference("com/opengamma/maths/datacontainers/ExprEnum", &_OGExprEnumClazz);
-  registerGlobalClassReference("com/opengamma/maths/datacontainers/matrix/OGSparseMatrix", &_OGSparseMatrixClazz);
-  registerGlobalClassReference("com/opengamma/maths/datacontainers/scalar/OGScalar", &_OGScalarClazz);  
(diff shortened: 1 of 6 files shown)


src/jshim/jvmmanager.cc
```diff
@@ -62,8 +62,10 @@ JVMManager::initialize(JavaVM *jvm)
 {
   VAL64BIT_PRINT("VM pointer", jvm);
 
+  JNIEnv * env;
+
   // Attempt to get env for JNI version 1.2
-  int ret = jvm->GetEnv((void **)&_env, JNI_VERSION_1_2);
+  int ret = jvm->GetEnv((void **)&env, JNI_VERSION_1_2);
```
gmarkall, line 68: The env that we're getting here looks unused - do we need this call to remain?

gmarkall, line 68: Summary of out-of-band conversation - we're leaving it here to check the JNI environment has the correct version available.

gmarkall:
This looks good. After some discussion:
- We think we can eliminate some calls to `JVMManager::GetEnv()` by passing the environment obtained with `jvm->GetEnv()` in `JVMManager::initialize()` to the `registerReferences()` function.
- This has the advantage that we're reducing the number of calls to `JVMManager::getEnv()` with the goal of eventually removing it in favour of passing the JIN environment pointer provided to us by the JVM in the JNI call, which should generally be safer.
- However, as there are a couple of use cases for `JVMManager::getEnv()` that are a little more fiddly to remove, and there may be a future need for this function, we're going to leave it in and remove at a later stage if no more use-cases are found.
- Generally, we don't expect to need to get an environment at points in the code where we don't already have one - the typical case where we may need one is for using for reporting an error back to the JVM, but we have a mechanism for doing this relatively cleanly with C++ exceptions and the C++->Java exception translation logic.

stuartarchibald:
Thanks for reviewing. In response to your comments, I have now refactored the offending code such that the `JVMManager::getEnv()` function is unused. In places where a `JNIEnv *` is needed the reference is obtained via passing the pointer through the function args. A review of the usefulness of the `JVMManager::getEnv()` function is slated for 6 month's time, ticket [[MAT-406]](http://jira.opengamma.com/browse/MAT-406).

Coverage of `src/jshim` drops 0.1% as a result of refactor.

gmarkall:
This looks great! Happy to merge.

PR #105, opened 2014-06-04
Use shared_ptr for managing the lifetime of OGNumeric objects

gmarkall:
## Summary

In summary, we wrap `OGNumeric` pointers inside a `shared_ptr` so
that we don't have to explicitly manage their deletion. This makes
it a lot easier to clean up during exceptions (and in general).

The form of DOGMA code is now changed to use typedef'd shared
pointers and factory methods. For example, where we had:

``` C++
const OGNumeric* my_num = new OGRealScalar(1);
```

before, we now have:

``` C++
OGNumeric::Ptr my_num = OGRealScalar::create(1);
```

There is a `::Ptr` type for every OGNumeric subclass. One could
similarly have declared `my_num` as:

``` C++
OGRealScalar::Ptr my_num;
```

Instances of these types can generally be passed around and used
in a similar manner to that in which a raw pointer would be used.

The factory methods are used to instantiate new objects. The
constructors have been made private, in order to prevent the use
of "naked" new and delete calls, which would defeat the purpose of
managing objects with `shared_ptr` (and would also have some other
implementation difficulties).
## PtrVectors

Since the `PtrVector` type was used to some extent for managing
the deletion of its contents, it is no longer required.
`RegContainer` and `ArgContainer` are now an `std::vector` of
`OGNumeric::Ptr` objects. As it is no longer required, `PtrVector`
is removed.
## Dispatch to terminals

The `Visitor` implementation of dispatch was a little tricky to
keep working during the conversion of the code to use
`shared_ptr`. In order to simplify it, the visitors for
dispatching to Java arrays of arrays were replaced by two other
classes, `Real16AoA` and `Complex16AoA` that dispatch using the
`OGNumeric` type identification instead of the visitor pattern.
This is a much simpler dispatch mechanism.

The dispatch mechanism that materialises to `OGTerminal`
instances is still in existence. However, it could be replaced
with a class similar to `Real16AoA`. Once this is done, I believe
that there would be no use for the visitor mechanism, since it
provides less flexible dispatch than using the `OGNumeric` type
identification, and usually requires a whole lot more code.
## Memory usage

Running the fuzzer for about an hour grew the memory usage up to
736MB resident, which remained fairly stable for the duration of
the run - a small increase of < 1MB per minute. The growth up
to 736MB may be largely accounted for by growing the tree cache.
The subsequent growth requires further investigation to quantify.

Note that I do not believe this proves that we are leak free -
however, it does provide a good indication that we are not leaking
horrendously.
## Further details

Further details of changes are in the commit logs - I have avoided
repeating them here in order to save having to read the same
changes twice.
## Test coverage

Java: 79%
build/src/librdag: 16.4% (up 0.3%)
src/jshim: 54.5% (down 3% due to shortness of
Real16AoA/Complex16AoA vs. the old DispatchToReal16ArrayOfArrays)
src/librdag: 91.4% (down 0.4%, due to removal of containers, and
shortening of entrypt)
src/librdag/runners: 95.7% (unchanged)


stuartarchibald:
**Manual check of leak status.**
Checked via:
- Running the fuzzer on this branch locally with `java -Xmx2G`
- Attaching `jvisualvm` to the process.
- Running `pidstat -r` on the PID.

First there was 10 minutes of transient behaviour as the JVM sorted itself out. The next 50 odd minutes ran fairly smoothly. According to `jvisualvm` the JVM heap increased in size by 10Mb over the stable period and `pidstat` reported a resident set memory increase at a rate of about 175Kb/minute ~= 10Mb/hr (from sample of (793116Kb - 790488Kb) / 15 minutes). I'm fairly confident with some more intelligent and long term profiling we can quantify this behaviour with greater accuracy. But in any case, without looking at the code in detail yet, these changes have really tightened up the memory management, so good job.

CMakeLists.txt
```diff
@@ -141,8 +141,15 @@ set(TARGET_TYPES dbg std sse41 avx1)
 # break comparison behaviours.
 set(FFAST_MATH_SAFE "-fno-math-errno  -fno-rounding-math -fno-signaling-nans -fcx-limited-range")
 
-set(OPT_FLAGS_dbg "-DDEBUG ${CMAKE_C_FLAGS_DEBUG}")
-set(OPT_FLAGS_std "-g ${FFAST_MATH_SAFE} -O2 -funroll-loops -DNDEBUG")
+if (WIN32)
+  # We need a larger than default stack to prevent the entrypt test
+  # failing, and we need to force the export of all symbols since
+  # they were all visible by default in the original implementation.
+  set (PLATFORM_FLAGS "-Wl,--export-all-symbols -Wl,--stack,4194304")
```
stuartarchibald, line 148: 4Mb, magic number or matched to *nix variants?

gmarkall, line 148: This is a magic number - the default of 1MB caused the tests to fail. The next I tried was 4MB, which allows tests to pass, so I left it there.

stuartarchibald, line 148: Ok. I'm happy to leave this as is given we have a record of it being magic.

src/generator/dispatch.py
```diff
@@ -184,12 +181,10 @@ def method_definitions(self):
                 else:
                     d = { 'argno': '1', 'typetoconvertto': type_to_convert_to }
                     conv1 = dispatchbinaryop_conv_arg %d
-                    deletions += dispatchbinaryop_deletion % { 'argno': '1' }
                 d = { 'node0type': t0.typename, 'node0enumtype': t0.enumname,
                       'node1type': t1.typename, 'node1enumtype': t1.enumname,
                       'conv0': conv0,
-                      'conv1': conv1,
-                      'deletions': deletions }
+                      'conv1': conv1 }
```
stuartarchibald, line 187: Am pleased these changes make the autogenned implementations so much easier to write too :)

src/jshim/jdispatch.cc
```diff
@@ -17,202 +17,204 @@ namespace convert
...
+  case REAL_MATRIX_ENUM:
+  case REAL_DIAGONAL_MATRIX_ENUM:
+  case REAL_SPARSE_MATRIX_ENUM:
+    _data = node->asOGTerminal()->toReal16ArrayOfArrays();
+    _rows = node->asOGTerminal()->getRows();
+    _cols = node->asOGTerminal()->getCols();
+    break;
+  default:
+    stringstream message;
+    message << "Unknown type for dispatchToReal16AoA. Type is " << type << ".";
```
stuartarchibald: Another case set is needed here, default to "unknown" is fine but in complex cases we know the type, its just the operation is not necessarily applicable. As a separate issue, we also need to have a think about type promotion, e.g. if a complex space terminal has a zero imaginary part should we promote to equivalent real type and then convert?

gmarkall: The old error message would just have said something like "DispatchToReal16ArrayOfArrays::visit(librdag::OGScalar<complex16>)" - I believe I have the same behaviour here, but it's just that the message is possibly misleading. Perhaps "Unsupported type for..." would be more appropriate?

I agree that we should have a think about type promotion in this case as well.

stuartarchibald: Ok. Let's leave it for now as it's not technically a regression.
[[MAT-408]](http://jira.opengamma.com/browse/MAT-408) is to fix up these messages, ideally we want "unsupported" and "unknown" cases.
[[MAT-409]](http://jira.opengamma.com/browse/MAT-409) is to discuss type promotion options.

src/jshim/jdispatch.cc
```diff
@@ -17,202 +17,204 @@ namespace convert
...
+  case COMPLEX_MATRIX_ENUM:
+  case COMPLEX_DIAGONAL_MATRIX_ENUM:
+  case COMPLEX_SPARSE_MATRIX_ENUM:
+    _data = node->asOGTerminal()->toComplex16ArrayOfArrays();
+    _rows = node->asOGTerminal()->getRows();
+    _cols = node->asOGTerminal()->getCols();
+    break;
+  default:
+    stringstream message;
+    message << "Unknown type for dispatchToReal16AoA. Type is " << type << ".";
```
stuartarchibald: This message needs fixing, given it refers to the wrong function!

gmarkall: Yes. I might as well correct the wording of the other one at the same time.

src/jshim/jshim.cc
```diff
@@ -386,8 +373,6 @@ Java_com_opengamma_maths_materialisers_Materialisers_materialiseToOGTerminal(JNI
     return nullptr;
   }
 
-  delete chain;
-
```
stuartarchibald, line 390: This set of functions is a lot cleaner now :)

src/jshim/test/check_jdispatch.cc
```diff
@@ -352,17 +294,14 @@ TEST(JDispatch, Test_DispatchToComplex16ArrayOfArrays_OGComplexSparseMatrix)
...
   int * rowInd=new int[5]{0, 2, 0, 1, 3};
   int * colPtr=new int[4]{0, 2, 4, 5};
 
-  OGRealSparseMatrix * mat = new OGRealSparseMatrix(colPtr, rowInd, data, rows, cols);
-  ASSERT_ANY_THROW(d->visit(mat));
-  delete d;
-  delete mat;
+  OGNumeric::Ptr mat = OGRealSparseMatrix::create(colPtr, rowInd, data, rows, cols);
+  ASSERT_ANY_THROW(Complex16AoA{mat});
   delete[] data;
```
stuartarchibald, line 305: I like how these changes have made implementing test code considerably less error prone.

src/jshim/test/check_jterminals.cc
```diff
@@ -146,13 +150,15 @@ TEST(JTerminals, Test_JOGIntegerScalar_ctor)
...
+    JOGIntegerScalar::Ptr scalar = JOGIntegerScalar::create(obj);
     ASSERT_TRUE(scalar->getValue()==value);
 
     // debug print for coverage purposes
     scalar->debug_print();
 
-    delete scalar;
+    // Need to delete the node now so that the fake environment and jvm are still around
+    // for its unbinding
+    scalar.reset();
```
stuartarchibald, line 161: Yes! Good point.

src/librdag/terminal.cc
```diff
@@ -47,22 +47,22 @@ OGTerminal::toComplex16ArrayOfArrays() const
...
+OGTerminal::Ptr
 OGTerminal::asOGTerminal() const
 {
-  return this;
+  return static_pointer_cast<const OGTerminal, const OGNumeric>(shared_from_this());
 }
 
 bool
-OGTerminal::operator==(const OGTerminal&  other) const
+OGTerminal::operator==(const OGTerminal::Ptr& other) const
```
stuartarchibald, line 57: We've been over the issues this causes to syntax off line.
[[MAT-410]](http://jira.opengamma.com/browse/MAT-410) to fix.

gmarkall:
A note from our discussion: the use of FuzzyCompareOGTerminalContainer is quite unclean. I've added [[MAT-411]](http://jira.opengamma.com/browse/MAT-411) for fixing it.

src/librdag/test/check_entrypt.cc
```diff
@@ -40,15 +38,13 @@ typedef pair<const OGNumeric*, const OGNumeric*> TreeResultPair;
...
-  delete node;
-  delete expectedResult;
-  delete result;
-
+  OGNumeric::Ptr node = param.first;
+  OGTerminal::Ptr expectedResult = param.second;
+  OGTerminal::Ptr result = entrypt(node);
+  ASSERT_NE(result, OGNumeric::Ptr{});
+  // PR: There is something very unexpected about this syntax.
+  // PR This is super-confusing! Needs API fixing
```
stuartarchibald, line 46: Noted in [[MAT-410]](http://jira.opengamma.com/browse/MAT-410)

gmarkall:
I've made the suggested amendments to jbindings.cc - waiting on the buildbot pass.

stuartarchibald:
**Final Comments**
1. First, well done. A heroic effort! Amazed it took only the time it did given the churn. Think this is the biggest and best design alteration since the initial import, so props to you for getting it done.
2. My review of this is _best effort_, the churn is understandably large and the compiler has a larger stack than my brain when it comes to errors. I've been looking largely to ensure that translation from the old raw pointers to the new encapsulating pointers is valid, and the method by which this is done is appropriate. I've also been looking at implementation details of core the parts to try and fathom if there is any problem in the logic. Due to the reasonably high level of coverage from the unit tests I feel we can gain additional confidence that this refactor is correct in that the coverage is similar and tests all pass.
3. I don't really have anything specific to say regarding the implementation (other than the tickets added inline) because this has been discussed as work progressed. Further, the restrictions on the `shared_ptr<>` usage forces certain behaviours such that there's only really one sensible way to do certain things and that's what's been implemented.
4. Of the comments above, there's a couple of minor things to patch, which appears to have now passed CI. The rest get tickets as this unit of work needs wrapping up.
5. As noted earlier the memory usage now appears approximately stable, which is fantastic news.

More than happy to merge!

PR #107, opened 2014-06-05
Fix FuzzyCompareOGTerminalContainer and ConvertTo interfaces in OGTerminal

gmarkall:
This PR involves:
- Changing the interface of `FuzzyCompareOGTerminalContainer` in `OGTerminal` so that no dangling references can be exposed, and `OGTerminal` no longer holds a reference to a `FuzzyCompareOGTerminalContainer` instance. This allows `FuzzyCompareOGTerminalContainer` to hold a `shared_ptr` instead of a `weak_ptr`, which will also prevent it from holding an expired pointer.
- Removing external access to the `ConvertTo` that has held by `OGTerminal`, as it is not needed externally, and would allow a dangling pointer. 

Coverage:

Java: 79%
build/src/jshim: 0.8%
build/src/librdag: 16.4%
src/jshim: 55%
src/librdag: 91.3%
src/librdag/runners: 95.7%

All unchanged except for src/librdag, which is down 0.1% because of the deletion of code that was covered.

### DIFF
--- a/include/convertto.hh
+++ b/include/convertto.hh
@@ -6,7 +6,8 @@
 
 #ifndef _CONVERTTO_HH
 #define _CONVERTTO_HH
-#include "terminal.hh"
+
+#include <memory>
 
 // defines permissable type conversions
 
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Thanks for this patch, I've not liked the return-ref since the day it was written (full well knowing the zillion line refactor needed [https://github.com/OpenGamma/OG-Maths/pull/105] to get the right pointer encapsulation to do this bit right, and consequently putting it off!). Happy to merge.

PR #108, opened 2014-06-05
[MAT-343] Always specify expected exceptions

gmarkall:
The use of ASSERT_ANY_THROW/EXPECT_ANY_THROW allows for things to
go wrong that we're not expecting, and still get a test pass. This
commit changes all ASSERT_ANY_THROW and EXPECT_ANY_THROW for
ASSERT_THROW and EXPECT_THROW.

Additionally, after doing this it was discovered that conversion
of real diagonal and real sparse matrices to Complex16AoA was not
actually working, and was throwing an rdag_error. Therefore,
support for attempting to do this has been removed from
Complex16AoA. I have created [[MAT-413]](http://jira.opengamma.com/browse/MAT-413) for this.

Coverage: unchanged.

### DIFF
--- a/src/jshim/jdispatch.cc
+++ b/src/jshim/jdispatch.cc
@@ -120,9 +120,7 @@ Complex16AoA::Complex16AoA(const OGNumeric::Ptr& node)
     break;
   case REAL_MATRIX_ENUM:
   case COMPLEX_MATRIX_ENUM:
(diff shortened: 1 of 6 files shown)


stuartarchibald:
Being more specific is a good idea, I believe all `{ASSERT/EXPECT}_*_THROW` statements are now of the more specific variety. I'm happy with these changes so am merging.

PR #110, opened 2014-06-05
Delete local refs of expression nodes as soon as possible.

gmarkall:
In order to reduce the local ref requirements, we can delete local refs of expression nodes once they are translated. We cannot do the same for local refs of terminals, because they are needed for unbinding arrays when we are cleaning up.

This commit achieves the deletion of local refs for expression nodes by merging the stages of the tree translation (which is enabled to be done cleanly by the new memory management) and deleting local refs of expression nodes as they are popped off the stack of working expression nodes.

All coverage unchanged, except for build/jshim, which is up to 54.1% from 53.7%. This is because the `createExpression` function is not tested from the C++ side, and is shortened by these commits.

### DIFF
--- a/src/jshim/exprfactory.cc
+++ b/src/jshim/exprfactory.cc
@@ -91,19 +91,14 @@ OGNumeric::Ptr createExpression(jobject jexpr)
   JNIEnv* env = nullptr;
   JVMManager::getEnv((void **) &env);
 
-  // Stage 1: linearise java tree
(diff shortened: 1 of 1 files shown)


stuartarchibald:
I'm happy with these changes. Keeps the local ref count down and puts less strain on the jvm. It's also now trivial to alter this behaviour if all the jvm queries start to impact performance as it's a single line to comment! Merging.

PR #112, opened 2014-06-10
Add logging using SLF4J

gmarkall:
In order to avoid a dependency on SLF4J being created, the `Logger`
class provides a logging interface for OG-Maths. It checks to see
if SLF4J is available - if it is, then new instances of it use
SLF4J. Otherwise, we fall back to a simple implementation that
prints all log messages to stderr.

Testing is added, but is limited to instantiating each of the
logging classes and attempting to use each of their methods.

An extra test is added that runs with all the JARs on the
classpath, so that NativeLibrariesSelfTest gets run both with and
without SLF4J available. Some instances of NativeLibrariesSelfTest
will still fail if we accidentally introduce a hard dependency on
SLF4J.

Calls to `println` are generally replaced with calls to the `Logger`.
This makes the library quiet from the Java side. Some `println`
calls remain, in places where the user would expect to see output
- for example, in the fuzzer and NativeLibrariesSelfTest.

Java coverage: 81% (unchanged)
Native coverage unchanged.

### DIFF
--- a/src/java/pom.xml
+++ b/src/java/pom.xml
@@ -21,5 +21,10 @@
 						<artifactId>testng</artifactId>
 						<scope>provided</scope>
 				</dependency>
+				<dependency>
+						<groupId>org.slf4j</groupId>
+						<artifactId>slf4j-api</artifactId>
+						<version>1.7.5</version>
+				</dependency>
 		</dependencies>
 </project>
--- a/src/java/src/main/CMakeLists.txt
+++ b/src/java/src/main/CMakeLists.txt
@@ -99,7 +99,18 @@ add_jar(${JAR_NAME_NO_EXTN} ${JAR_SOURCES}
 
 foreach(TARGET ${TARGET_TYPES})
   if(SUPPORT_${TARGET})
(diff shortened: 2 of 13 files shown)


src/java/pom.xml
```diff
@@ -21,5 +21,10 @@
 						<artifactId>testng</artifactId>
 						<scope>provided</scope>
 				</dependency>
+				<dependency>
+						<groupId>org.slf4j</groupId>
+						<artifactId>slf4j-api</artifactId>
+						<version>1.7.5</version>
```
stuartarchibald, line 27: Is this latest or just what we have in the various build systems? http://www.slf4j.org/download.html suggests 1.7.7 ?

gmarkall, line 27: It's what we have in various build systems.

stuartarchibald, line 27: Fair enough.

stuartarchibald:
Thanks for the later patches to pass the class name through to the stderr logger impl so a reference to the logging class appears in the message. I'm happy with this code. Merging.

PR #113, opened 2014-06-12
Rename real types to correct width

gmarkall:
This pull request involves:
- Renaming real8 to real4
- Renaming real16 to real8
- Eliminates the use of double and complex<double> where we should
  be using real8 and complex16

Please see commit logs for individual substitution command lines.

Coverage: unchanged.


src/librdag/test/check_entrypt.cc
```diff
@@ -98,7 +98,7 @@ TreeResultPair negatecomplexmatrix()
 
 TreeResultPair negates[] =
    { negatepair(1.0), negatepair(-1.0), negatepair(0.0),
-     doublenegate(1.0), doublenegate(-1.0), doublenegate(0.0),
+     real8negate(1.0), real8negate(-1.0), real8negate(0.0),
```
stuartarchibald: Contextually invalid substitution! Please patch back to original.

gmarkall: Ah, yes, I see. Will patch back.

stuartarchibald:
A few minor things were picked out over IRC where `sed` had chomped something that had other meaning. Few inconsistencies in variables names now sorted too. I believe the refactor is clean and now names are consistent with the type system. Thanks for the patches. Happy to merge.

PR #114, opened 2014-06-12
Remove `using namespace std;` from header files

gmarkall:
It is a bit rude to include the whole of the `std` namespace in our header files, because anyone including our headers then gets the whole of `std` in their global namespace.

This PR removes all `using namespace std;` from headers, mainly by pushing it down in to the source files. There are some instances where it was straightforward to use exactly the members of `std` that were required - in some cases I have done this.

Coverage: unchanged.

### DIFF
--- a/include/exceptions.hh
+++ b/include/exceptions.hh
@@ -17,8 +17,6 @@
  */
 #define FRAMEBUF_SIZE 50
 
-using namespace std;
-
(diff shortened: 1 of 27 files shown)


stuartarchibald:
I'm happy with this change, much neater approach. Thanks for the patch. Merging.

PR #115, opened 2014-06-16
Use int4 and size_t where appropriate

gmarkall:
This PR replaces the `int` type with either `size_t` or `int4` in most cases. `int4` is an `int32_t`. 

Generally, `size_t` is used for anything that is the size of an array or count of elements. Every other integer is an `int4`. The exceptions to this are:
- Lapack is fortran and is therefore using signed `int` types - anything that touches Lapack uses an `int4` for sizes instead of `size_t`.
- The JVM typedefs `jint` to be `int` on Unix-like platforms and `long int` on Windows. It is therefore necessary to keep `jint` in some places.
- Some library functions (e.g. `abi::__cxa_demangle`) expect a pointer to an `int` to return their status in, so we use `int` for them.

Supporting these changes also involves:
- `getIntFromVoidJMethod` has been replaced with `getSizeTFromVoidJMethod`. This checks that the value it gets from Java is non-negative and casts it to a `size_t`.
- The checks for negative sizes of things in the terminals were removed, since negative sizes are no longer possible with `size_t`. Values coming in from Java are still check for non-negativity, because  `getSizeTFromVoidJMethod` checks them.
- Most of the runner implementations get the number of rows and columns from terminals, which are of type `size_t`. However, because they are passed to Lapack, `int4` types are used to hold the values that are passed to Lapack.

This PR also includes the removal of `PrintTreeVisitor`, which was dead code.

Coverage:

Java: unchanged.
Native: unchanged, except for:
src/jshim: 54.3%, up from 54.1%
src/librdag: 94.8%, up from 91% (due to deletion of PrintTreeVisitor)


stuartarchibald:
Thanks for the patch. To aid with review, any chance you could please paste in the output of something like `egrep -R '\bint\b' ${OG_MATHS_ROOT}` with view of checking the validity of remaining `int` locations?

gmarkall:
I'm avoiding the `src/java` and `external` folders since they will have lots of `int`s in there that we don't want to touch:

```
$ grep "\bint\b" include/ src/librdag/ src/jshim/ -R
include/jbindings.hh:// We must use jint instead of int4 because the definition of jint varies between int and long int
include/jbindings.hh:// We must use jint instead of int4 because the definition of jint varies between int and long int
include/jbindings.hh:// We must use jint instead of int4 because the definition of jint varies between int and long int
include/debug.h:    low = (long long unsigned int)val & 0x00000000FFFFFFFFULL; \
include/debug.h:    high = ((long long unsigned int)val & 0xFFFFFFFF00000000ULL) >> 32
include/debug.h:  unsigned int SUPPRESS_UNUSED __ptr_print_high, SUPPRESS_UNUSED __ptr_print_low; \
include/debug.h:  DEBUG_PRINT(desc " is 0x%llx\n", (long long unsigned int)ptr);
include/test/fake_jvm.hh:    virtual int GetEnv(void ** env, int SUPPRESS_UNUSED version) override
include/fake_jni.h:    virtual int* GetIntArrayElements(jintArray SUPPRESS_UNUSED arr, bool SUPPRESS_UNUSED *isCopy)
include/fake_jni.h:    virtual void ReleaseDoubleArrayElements(jdoubleArray SUPPRESS_UNUSED arr, double SUPPRESS_UNUSED *nativeArr, int SUPPRESS_UNUSED mode)
include/fake_jni.h:    virtual void ReleaseIntArrayElements(jintArray SUPPRESS_UNUSED arr, int SUPPRESS_UNUSED *nativeArr, int SUPPRESS_UNUSED mode)
include/fake_jni.h:        virtual int GetEnv(void ** env, int version) = 0;
include/fake_jni.h:    int _version;
include/fake_jni.h:typedef int jsize;
include/jterminals.hh: * helper function to get size_ts from int getFOO() in java
src/librdag/mem.cc:  int ret = posix_memalign((void**)&memptr, __ALIGNMENT, size);
src/librdag/exceptions.cc:// status is int as abi::__cxa_demangle status is int
src/librdag/exceptions.cc:const string demangle_fname(const string& fname, int* status)
src/librdag/exceptions.cc:  int status = 0;
src/librdag/numerictypes.cc:    unsigned long long int i;
src/librdag/numerictypes.cc:    unsigned long long int i;
src/librdag/numerictypes.cc:    unsigned long long int i;
src/librdag/numerictypes.cc:    unsigned long long int i;
src/jshim/test/check_jvmmanager_fakejni.cc:        virtual int GetEnv(void SUPPRESS_UNUSED ** env, int SUPPRESS_UNUSED version) override
src/jshim/test/check_jvmmanager_fakejni.cc:        _jobjectArrayImpl(int len)
src/jshim/test/check_jvmmanager_fakejni.cc:        int getLen()
src/jshim/test/check_jvmmanager_fakejni.cc:        int _len;      
src/jshim/test/check_jvmmanager_fakejni.cc:        _jdoubleArrayImpl(int len)
src/jshim/test/check_jvmmanager_fakejni.cc:        int getLen()
src/jshim/test/check_jvmmanager_fakejni.cc:        int _len;      
src/jshim/test/check_jvmmanager_fakejni.cc:      virtual int GetEnv(void SUPPRESS_UNUSED ** env, int SUPPRESS_UNUSED version) override
src/jshim/test/check_jterminals.cc:    virtual int * GetIntArrayElements(jintArray SUPPRESS_UNUSED arr, bool  SUPPRESS_UNUSED *isCopy) override
src/jshim/test/check_jterminals.cc:      return reinterpret_cast<int *>(&_value);
src/jshim/test/check_jterminals.cc:    int value = 1234;
src/jshim/test/check_jterminals.cc:    Fake_JNIEnv * env  = new Fake_JNIEnv_for_binding<int>(value);
src/jshim/test/check_jterminals.cc:    Fake_JNIEnv_for_OGSparseMatrix_T(size_t rows, T * value, int * idx):Fake_JNIEnv_for_OGMatrix_T<T>(rows,value)
src/jshim/test/check_jterminals.cc:    int * _idx;
src/jshim/test/check_jterminals.cc:    int * idx = new int[4]{1,2,2,2};
src/jshim/test/check_jterminals.cc:    int * idx = new int[4]{1,2,2,2};
src/jshim/jterminals.cc: * helper function to get size_ts from int getFOO() in java
src/jshim/jvmmanager.cc:  int ret = jvm->GetEnv((void **)&env, JNI_VERSION_1_2);
```

src/librdag/exceptions.cc
```diff
@@ -107,6 +107,7 @@ string get_fname(const string& backtrace_string)
   return fname;
 }
 
+// status is int as abi::__cxa_demangle status is int
```
stuartarchibald, line 110: Confirmed.

stuartarchibald:
Adding tickets:
- [[MAT-434]](http://jira.opengamma.com/browse/MAT-434) for exception thrown on on `rows*cols > MAX_INT` in dense matrix case to prevent overflows when the native memory allocation comes into force.
- [[MAT-425]](http://jira.opengamma.com/browse/MAT-435) for using C++ `std::size_t` in preference to `size_t` from C's `types.h`.

stuartarchibald:
From a quick scan, I agree with the remaining items that are `int`, seem largely bound to decisions in `jni.h`.

stuartarchibald:
My human parse of this can't see any case where a bound could be a `size_t` where it isn't and where an `int4` ought to be a `size_t`. I agree with the latest-possible conversion of `size_t` to `int4` in cases where needed (runners). The use of `int4` for `colPtr` and `rowIdx` in the sparse types is valid at present due to the libraries that do the lifting typically using `int32_t` internally. I'm given confidence by the build being clean, however, I'm adding [[MAT-436]](http://jira.opengamma.com/browse/MAT-436) to add in tests that use huge data which is likely to trip overflows, perhaps the fuzzer API could be employed for such tests? Anyway, I'm happy with this patch and so am merging. Many thanks.

PR #116, opened 2014-06-17
Mat 355

stuartarchibald:
Ticket [[MAT-355]](http://jira.opengamma.com/browse/MAT-355) is to write the super solver, this PR is part way there and is with view of adding in the LAPACK functions needed to do the heavy lifting. It:
- Adds a load more functions from LAPACK in templated form.
- Switches to use `unique_ptr<>` for memory management for templated LAPACK internals.
- Does the double rebase on type changes and incorporates new `real8` and `int4` types.
- Adds end to end tests to make sure the templates fire and clean up correctly.
- Fixes a bug in error reporting, bad arg is always at `-(*INFO)`.

The patches look at bit messy and fiddly, partly because of the rebase, partly because of the switch to `unique_ptr<>` and partly because LAPACK is fiddly, apologies!

**Coverage**
Java: (no change)
build/include 58.7% (no change)
build/src/jshim     0.8% (no change)
build/src/librdag 16.4% (no change)
src/jshim   54.3% (no change)
src/librdag 94.3% (down 0.5%, exceptions in LAPACK templates for things like failed convergence not covered).
src/librdag/runners     95.7% (no change)
src/librdag/test/nodes 96.9% (no change)

### DIFF
--- a/include/lapack.hh
+++ b/include/lapack.hh
@@ -29,6 +29,7 @@ extern char U;
 extern char D;
 extern char ONE;
 extern char O;
+extern char V;
 extern int4 ione;
 extern int4 izero;
 extern real8 rone;
@@ -37,6 +38,8 @@ extern real8 rzero;
 extern complex16 czero;
 
 template<typename T>char charmagic();
+template<typename T>std::string xxxgqrcharmagic();
+
 
 template<typename T> void xscal(int4 * N, T * DA, T * DX, int4 * INCX);
 template<typename T>void xswap(int4 * N, T * DX, int4 * INCX, T * DY, int4 * INCY);
@@ -49,6 +52,12 @@ template<typename T> void xpotrf(char * UPLO, int4 * N, T * A, int4 * LDA, int4 
 template<typename T> void xtrtrs(char * UPLO, char * TRANS, char * DIAG, int4 * N, int4 * NRHS, T * A, int4 * LDA, T * B, int4 * LDB, int4 * INFO);
 template<typename T> void xgetrf(int4 * M, int4 * N, T * A, int4 * LDA, int4 * IPIV, int4 *INFO);
 template<typename T> void xgetri(int4 * N, T * A, int4 * LDA, int4 * IPIV, T * WORK, int4 * LWORK, int4 * INFO );
+template<typename T> real8 xlange(char * NORM, int4 * M, int4 * N, T * A, int4 * LDA, real8 * WORK );
+template<typename T> void xgetrs(char * TRANS, int4 * N, int4 * NRHS, T * A, int4 * LDA, int4 * IPIV, T * B, int4 * LDB, int4 * INFO);
+template<typename T> void xgels(char * TRANS, int4 * M, int4 * N, int4 * NRHS, T * A, int4 * LDA, T * B, int4 * LDB, T * WORK, int4 * LWORK, int4 * INFO );
+template<typename T> void xgeqrf(int4 * M, int4 * N, T * A, int4 * LDA, T * TAU, T * WORK, int4 * LWORK, int4 *INFO);
+template<typename T> void xxxgqr(int4 * M, int4 * N, int4 * K, T * A, int4 * LDA, T * TAU, T * WORK, int4 * LWORK, int4 * INFO);
+
 }
 
 /**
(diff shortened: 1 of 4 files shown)


include/lapack.hh
```diff
@@ -306,7 +319,114 @@ real8 zlanhe(char * NORM, char * UPLO, int4 * N, complex16 * A, int4 * LDA);
...
+ * @param M as LAPACK dgelsd M.
+ * @param N as LAPACK dgelsd N.
+ * @param NRHS as LAPACK dgelsd NRHS.
+ * @param A data type specific with intent as LAPACK dgelsd A.
+ * @param LDA as LAPACK dgelsd LDA.
+ * @param B data type specific with intent as LAPACK dgelsd TRANS.
+ * @param LDB as LAPACK dgelsd LDB.
+ * @param S as LAPACK dgelsd S.
+ * @param RCOND as LAPACK dgelsd RCOND.
+ * @param INFO as LAPACK dgels INFO.
```
gmarkall, line 385: Missing RANK parameter

src/librdag/test/check_lapack.cc
```diff
@@ -743,3 +742,550 @@ TEST(LAPACKTest_xpotrs, zpotrs) {
...
   delete [] RHS;
 }
+
+
+TEST(LAPACKTest_xlange, dlange) {
+  int4 n = 4;
+  real8 * A = new real8[16];
+  std::copy(rspd,rspd+n*n,A);
+  real8 answer = lapack::xlange(lapack::O, &n, &n, A, &n);
+  EXPECT_TRUE(answer == 79);
```
gmarkall: Is this exact equality check safe?

src/librdag/test/check_lapack.cc
```diff
@@ -743,3 +742,550 @@ TEST(LAPACKTest_xpotrs, zpotrs) {
...
+  const int4 minmn = m > n ? n : m;
+  int4 INFO = 0;
+
+  real8 * Acpy = new real8[20];
+  std::copy(rcondok5x4,rcondok5x4+m*n,Acpy);
+
+  // need a 1 norm
+  real8 anorm = lapack::xlange(lapack::O, &m, &n, Acpy, &m);
+  // check it
+  EXPECT_TRUE(anorm==82.e0);
```
gmarkall: Another exact equality check?

src/librdag/test/check_lapack.cc
```diff
@@ -743,3 +742,550 @@ TEST(LAPACKTest_xpotrs, zpotrs) {
...
+  int4 two = 2;
+  lapack::xgelsd(&m, &n, &two, Acpy, &m, rhs, &m, S, &RCOND, &RANK, &INFO);
+
+  // answers are striped in "n" length columns in RHS
+  EXPECT_TRUE(ArrayFuzzyEquals(expected,rhs,n,1e-13,1e-13));
+  EXPECT_TRUE(ArrayFuzzyEquals(&expected[4],&rhs[5],n,1e-13,1e-13));
+
+  EXPECT_TRUE(ArrayFuzzyEquals(expected_S,S,minmn,1e-13,1e-13));
+
+  EXPECT_TRUE(RANK==n);
```
gmarkall: If you are fixing things up you may want to replace this with `EXPECT_EQ(RANK, n)` so that GTest will provide more helpful output if this expectation fails.

src/librdag/test/check_lapack.cc
```diff
@@ -743,3 +742,550 @@ TEST(LAPACKTest_xpotrs, zpotrs) {
...
```
gmarkall: `EXPECT_EQ` as above

gmarkall:
This looks great. The use of `unique_ptr` in `lapack.cc` is quite clever, even though you need a raw pointer to pass to lapack - it makes ensuring cleanup easier. The coverage looks good - the uncovered parts seem to be those cases where convergence would fail, which is acceptable.

There are one or two minor comments on the diff, but I'm happy to merge this after those fixups.

stuartarchibald:
Thanks for the review. I agree with your comments and will create a patch.

stuartarchibald:
Patch 38b725e attempts to address comments.

gmarkall:
Thanks for the patch - it looks great.

Happy to merge.

PR #117, opened 2014-06-18
Replace DispatchToOGTerminal with JavaTerminal and remove the visitor

gmarkall:
`DispatchToOGTerminal` is replaced with `JavaTerminal` so that we no longer use the visitor implementation. `JavaTerminal` uses our type dispatch mechanism in a similar way to `Real16AoA` and `Complex16AoA`.

This allows us to erase the visitor implementation that remained, since there is nothing left using it.

Coverage:

Java: unchanged.
src/jshim: 52.9%, down from 54%, because `DispatchToOGTerminal` and `JavaTerminal` have no C++ tests, and `JavaTerminal` is slightly longer.
src/librdag: 94.3% down from 94.4%, due to the removal of the accept methods which were tested.

All others unchanged.


src/jshim/jdispatch.cc
```diff
@@ -252,90 +289,77 @@ jdoubleArray extractComplexPartOfComplex16Arr2JDoubleArr(JNIEnv* env, complex16*
...
 void
-DispatchToOGTerminal::visit(librdag::OGMatrix<real8> const *thing)
+JavaTerminal::createRealMatrix(JNIEnv* env, const OGNumeric::Ptr& node)
 {
   jclass cls = JVMManager::getOGRealDenseMatrixClazz();
   jmethodID constructor = JVMManager::getOGRealDenseMatrixClazz_init();
-  jobjectArray darr = Real8AoA{OGNumeric::Ptr{thing->createOwningCopy()}}.toJDoubleAoA(_env);
-  jobject newobject = _env->NewObject(cls, constructor, darr);
-  setObject(newobject);
+  jobjectArray darr = Real8AoA{node}.toJDoubleAoA(env);
```
stuartarchibald, line 323: Nice :)

src/jshim/jdispatch.cc
```diff
@@ -349,77 +373,71 @@ jint* makeCJintArray(const int4* arr, size_t len)
...
   delete[] colPtr;
 
   // Row index
-  size_t datalen = thing->getDatalen();
-  jintArray jRowIdx = JVMManager::newIntArray(_env, datalen);
-  jint* rowIdx = makeCJintArray(thing->getRowIdx(), datalen);
-  _env->SetIntArrayRegion(jRowIdx, 0, datalen, rowIdx);
+  size_t datalen = mat->getDatalen();
+  jintArray jRowIdx = JVMManager::newIntArray(env, datalen);
+  jint* rowIdx = makeCJintArray(mat->getRowIdx(), datalen);
```
stuartarchibald: Perhaps this could be made to return a `unique_ptr<>` or similar managed memory type? Whilst not under the remit of this ticket, it is a relatively minor change (given the present churn) that I think would tighten the whole code up in the viscinity (same comment applies to the similar methods in here). If you are happy to do this it would be much appreciated, thanks.

stuartarchibald:
Thanks for this patch set, I feel it does the required and considerably neatens the code surrounding the creation of new terminals. Given we have RTTI out of necessity due to needing type identification to cross the DSO boundary, visitors are a somewhat pointless legacy that I'm glad to see gone. Many thanks for the fix ups regarding memory management. Happy to merge.

PR #118, opened 2014-06-19
Mat 420 rename dense matrix types to match java API

stuartarchibald:
This PR:
- Renames the C++ `OG{Real,Complex}Matrix` type to `OG{Real,Complex}DenseMatrix` to be consistent with the java API and also to be truer in name to the underlying data.
- Adjusts the enum names used in RTTI to match the change above.

The commit messages contain the expressions used to do most of the work. Some manual patching up was needed (e.g. autogeneration scheme).

No change in coverage as no additional logic was added, this is purely a variable name refactor.


gmarkall:
Many thanks for this - the existing changes look good. There are also some mentions of "RealMatrix" and "ComplexMatrix" that remain:

In jdispatch.hh/jdispatch.cc - `JavaTerminal` docstrings still refer to `OGRealMatrix` and `OGComplexMatrix`. The methods are also still called `createRealMatrix` and `createComplexMatrix` - perhaps these should be renamed to `createRealDenseMatrix` etc.? Additionally, there is an unrelated typo in most of these docstrings from the last PR (`@parame` instead of `@param`) which could do with fixing up if you are inclined to do so at the same time.

Less important mentions, since they're just part of test code (I don't mind whether these remain):

In check_svd.cc: `TEST(SVDTests,CheckRealMatrix)` / `TEST(SVDTests,CheckComplexMatrix)`
In check_expressions.cc: `TEST(VirtualCopyTest, RealMatrix)` / `TEST(VirtualCopyTest, ComplexMatrix)`

stuartarchibald:
Thanks for the review. I agree with your comments and have implemented a further patch (29c9518) to address them.

gmarkall:
Thanks for the fixups. This now looks great, and it's good to have consistency between the Java and C++ sides now. Merging.

PR #119, opened 2014-07-03
Mat 437 - implement the mldivide operation

stuartarchibald:
This PR contains the code required to add the `mldivide` (matrix left division|magic backslash|super solver [depends who you ask]) algorithm. The code is a rough translation of the java variant of this from the OG-Maths prototype, but with both algorithmic changes and bug fixes added. Also, thanks to the massive templating effort, the complex domain variant of this algorithm came for free.

Unfortunately this PR is fairly heavy going, the algorithm is complicated in the first place, the memory management is hard work and the testing is fairly tricky. I've tried to add as much commentary regarding the alg and what the different branches etc. are doing inline. Similarly, the unit test pairs for `<real8, complex16>` each have a one liner stating their properties. 

Coverage from the unit tests is extensive and in some places overlapping, this is with intention of ensuring the right branches are tripped. I've also manually checked each unit test pair to make sure that both real and complex data sets in the pair trip the same branches in the alg. Obviously the problem with a super solver is that it just keeps going so optimisation/efficiency attempts had to be manually checked to ensure e.g. a triangular matrix was indeed hitting a triangular back solver opposed to e.g. a SVD based solver.

Just as a process note: I've left the debugging flag on for now, so calls to this routine are a bit noisy as the decision tree is dumped to `stdout`. I intend to turn this off once we've run the fuzzer on this for a good long while, given the complexity, there's potential for bugs.

Further to the work on `mldivide`, I've added construct from `std::initializer_list` to the `OGXXXDenseMatrix` types on the C++ side, this considerably neatens and simplifies setting up input data in testing.

On the back of this work, it would appear sensible to create a "unrecoverable" error type from which a `xerbla_error` could be derived such that throws from the `lapack::` namespace of this sort of error are not caught under the standard `rdag_error&` catches. Similarly a `nonconformance_error` type would also be beneficial and be in the the same hierarchy. Ticket [[MAT-440]](http://jira.opengamma.com/browse/MAT-440) is raised to address this issue.

**Coverage**
Java: 77% (up 1%)
build/src/librdag 17.1% (up 0.5%)
src/jshim 53.7% (no change)
src/librdag 94.2% (down 0.2%)
src/librdag/runners 94.5% (down 1.2%, missed branches for xerbla based fails and a branch I think is dead (see below)).

### DIFF
--- a/src/librdag/runners/mldividerunner.cc
+++ b/src/librdag/runners/mldividerunner.cc
@@ -1,0 +1,881 @@
+/**
+ * Copyright (C) 2014 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for licence.
+ *
+ */
+
+#include "dispatch.hh"
+#include "runners.hh"
+#include "expression.hh"
+#include "iss.hh"
+#include "terminal.hh"
+#include "uncopyable.hh"
+#include "lapack.hh"
+#include "equals.hh"
+
+#include <stdio.h>
+#include <complex>
+#include <sstream>
+
+using namespace std;
+
+/*
+ * Unit contains code for MLDIVIDE node runners
+ */
+
+namespace librdag
+{
+
+/**
+ * Local implementation details.
+ */
+namespace detail
+{
+
+/**
+ * Template to create a "real" number in a given domain.
+ */
+template<typename T> T ndm(real8 val) {}
+
+template<> real8 ndm(real8 val)
+{
+  return val;
+}
+
+template<> complex16 ndm(real8 val)
+{
+  return complex16(val, 0.e0);
+}
+
+/**
+ * Casts a value to a char
+ * @param the value to cast
+ * @return the char representation of the value
+ */
+template<typename T> char asChar(T value)
+{
+  return static_cast<char>(value);
+}
+
+// Auxiliary structs for holding intermediary guiding data
+
+
+// NOTE: The following enums are used in some cases to pass flags to lapack.
+// Ideally lapack:: members holding `char *` would be used, but these are
+// declared extern which conflicts with the constexpr needed in the following.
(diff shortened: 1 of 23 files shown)


src/librdag/runners/mldividerunner.cc
```diff
@@ -0,0 +1,978 @@
...
+    }
+
+    // Is the array1 triangular or permuted triangular, send probe and get back struct containing result
+    unique_ptr<detail::TriangularStruct> tptr = detail::isTriangular(data1, rows1, cols1);
+
+    // Set flags based on triangular probe
+    detail::UPLO UPLO = tptr->flagUPLO;
+    detail::UNITDIAG DIAG = tptr->flagDiag;
+
+    // Is this matrix layour some breed of triangular?
```
stuartarchibald: correct typo in comment

gmarkall:
This looks very good - thanks for making numerous comments and descriptions throughout the algorithm. Although I'm not particularly familiar with LAPACK, it was helpful enough that I could follow what the code was doing as execution proceeded through the mldivide runner. 

There are one or two small things that could be fixed up:
- I'm happy for the debug prints to stay in, but can we use `std::cerr` for them so that they can be filtered from normal output?
- There's a typo "inveral" instead of "internal" in a comment in mldividerunner.cc
- In `mldivide_dense_runner` there is a catch block on line 646. This rethrows when `info<0`, and `info==0` is the case handled by the next if block. However, if `info>0` we catch this exception and proceed into the block beginning `if (!singular)`. Whilst this will probably not result in any error condition, it would be good to explicitly handle this `info>0` case in or near the catch block.

stuartarchibald:
Many thanks for reviewing. I believe that 9a900ee addresses the comments raised.

gmarkall:
Thanks for fixing up - this looks great!

PR #120, opened 2014-07-08
Mat-441Fix segfault occuring in pinv()

stuartarchibald:
The fuzzer spotted a segfault in the `pinv()` function, turns out this was from an invalid loop induction setting a limit of type `size_t` to `-1`, this limit was then used as a upper bound to index an array and hence segfaulted!

To fix this:
- A branch to test for all zeros in the input matrix is added, this is quicker to handle than a SVD anyway and allows an early escape.
- The offending loop is now patched and a further branch is added should it be possible to escape early when all singular values are below the threshold (this branch is not tested as it is a safety net and I'm not wholly convinced it's even reachable!).
- The svd runner is also patched to use smart pointer managed memory.

Coverage:
Java: 77% (no change)
build/src/librdag 17.1% (no change)
src/jshim 53.1% (no change)
src/librdag 94.2% (no change)
src/librdag/runners 94.6% (up 0.1%).

### DIFF
--- a/src/librdag/runners/pinvrunner.cc
+++ b/src/librdag/runners/pinvrunner.cc
@@ -85,7 +85,7 @@ pinv_dense_runner(RegContainer& reg, shared_ptr<const OGMatrix<T>> arg)
     const size_t minmn = m > n ? n : m;
 
     // Perform SVD on a copy of the argument, since it will get destroyed
-    OGExpr::Ptr svd = SVD::create(arg);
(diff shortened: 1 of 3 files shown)


gmarkall:
This looks good - thanks for fixing up the redundant copy for the SVD and the redundant unique_ptrs (mentioned in OOB conversation). 

The tests look like they're testing both the "all zeros" case and the case where `lim` gets set to a negative value, so I think this PR is good to merge.

PR #121, opened 2014-07-08
[MAT-443] fix segfault in mldivide from "triangle" pass.

stuartarchibald:
The fuzzer picked up a segfault in mldivide for the data presented in the regression test included. Essentially, matrix with a row containing all zeros, which is a valid input, could cause a segfault due to the value -1 being a flag and it not being overwritten by a `rowStart` as there effectively isn't one in this case. So a lookup of index `size_t = -1` occurs and causes havoc. Patch basically checks indices are valid before attempting the check of permutation validity. Regression test added straight from fuzzer input that found the bug.

Coverage:
Java: 77% (no change)
build/src/librdag 17.1% (no change)
src/jshim 53.1% (no change)
src/librdag 94.2% (no change)
src/librdag/runners 94.8% (up 0.2%).

### DIFF
--- a/src/librdag/runners/mldividerunner.cc
+++ b/src/librdag/runners/mldividerunner.cc
@@ -219,8 +219,7 @@ void checkIfUpperTriangularPermuteRows(T * data, size_t rows, size_t cols,  Tria
 
   // This holds the logic for whether we've seen a valid row of a triangular matrix before or not.
   std::unique_ptr<bool[]> rowTagPtr (new bool[rows]);
-  bool * rowTag = rowTagPtr.get();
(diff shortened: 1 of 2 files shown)


stuartarchibald:
Process note: I ran the fuzzer from a build of this branch on a 16 core linux box with:
`java -Xmx8G -Xms6G -cp ${CLASSPATH}:og-maths-lnx-0.1.0-SNAPSHOT.jar 'com.opengamma.maths.fuzzer.FuzzerMain' -t 6000 -c 16  -p > log_fix.txt 2>&1`
for over an hour without catastrophic incident.

gmarkall:
This looks good - happy to merge!

PR #122, opened 2014-07-11
Mat 425 add in OG-Izy wiring.

stuartarchibald:
This PR is to be read in conjunction with https://github.com/OpenGamma/OG-Izy/pull/4: 
- It adds in the ability to link against OG-Izy.
- It adds an izy namespace and then demonstrates the pattern needed to add a single function binding from izy to the namespace, and then tests it.
- It adds the ability to accept multiple subprojects in `verinfo.py`. It then adds the necessary wiring to include the buildinfo yaml from `izy` into the OG-Maths master yaml.
- It updates the properties file to include the izy libraries in the jar.

Note: I'm not convinced of the API in the `izy` namespace as yet. I've simply added in the most obvious thing just to test that the code is wired up correctly and gives the correct numerical results. The `izy` namespace and functionality is subject to both autogeneration and multiple interfaces. IIRC the original OG-Maths had multiple interfaces to this library too, I expect the same will be required here.

### DIFF
--- a/cmake/verinfo.py
+++ b/cmake/verinfo.py
@@ -27,20 +27,22 @@ def jarname(version, suffix):
         suffix = '-%s' % suffix
     return 'jars/og-maths-%s-%s%s.jar' % (platform_code(), version, suffix)
 
-def get_subprojects(lapack_verinfo_file):
-    """Returns a list of verinfo data for the OG-Maths subprojects. At the
-    moment, the only OG-Maths subproject is OG-Lapack."""
-    with open(lapack_verinfo_file, 'r') as f:
-        lapack_verinfo = load(f, Loader=Loader)
-    if len(lapack_verinfo['platforms']) != 1 or lapack_verinfo['platforms'][0] != platform_code():
-        raise RuntimeError('OG-Lapack platform mismatch')
-    return [lapack_verinfo]
+def get_subprojects(*args):
+    """Returns a list of verinfo data for the OG-Maths subprojects."""
+    ret = list()
+    for val in args:
+      with open(val, 'r') as f:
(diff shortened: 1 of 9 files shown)


stuartarchibald:
Process note: I've patched the CI build config to handle this change (`-D...OG-Izy...location` needed) and will commit it when PR patches are considered satisfactory.

cmake/verinfo.py
```diff
@@ -27,20 +27,22 @@ def jarname(version, suffix):
...
-        raise RuntimeError('OG-Lapack platform mismatch')
-    return [lapack_verinfo]
+def get_subprojects(*args):
+    """Returns a list of verinfo data for the OG-Maths subprojects."""
+    ret = list()
+    for val in args:
+      with open(val, 'r') as f:
+          verinfo = load(f, Loader=Loader)
+      if len(verinfo['platforms']) != 1 or verinfo['platforms'][0] != platform_code():
+          raise RuntimeError('OG subproject platform mismatch')
```
gmarkall: Could you state which subproject mismatches in the message?

gmarkall:
I think this looks good. The interface to IZY looks sensible to me on the face of it - however, I agree that it is going to be sensible to generate that interface for all the functions that make it up. The test that ensures the existing `vx_cos` function is correctly hooked up looks sane to me.

There are a couple of things that would be nice to fix up:
- It might be helpful to report which subproject mismatches when a mismatch occurs. 
- Could you update INSTALL.md to describe the Izy dependency and how it changes the build process?

Also, I'll hold off merging this PR whilst we go through the corresponding libizy one, and get that merged first.

gmarkall:
Thanks for fixing those things up. I'm happy to merge this.

PR #123, opened 2014-07-14
[MAT-444] remove dynamic libstdc++

stuartarchibald:
In this PR:
The dynamic linking of `libstdc++` is swapped for static linking.

This is done, in the parlance of GCC, to "freeze" the version of `libstc++` against which the build links. This is so that C++ symbols are guaranteed to resolve correctly regardless of other `libstdc++` versions potentially loaded, which under dynamic linking circumstances may prevent the loading of our "correct" ones.

In reviewing, please, along with anything else you see fit, confirm that the built article:
- Does not dynamically link against `libstdc++`.
- Does run, with no modules loaded, on the HPC dev box which altered us to this issue in the first place.

Coverage:
No change as diff is in the build system/bootstrapping config only.


gmarkall:
I've checked that the linkage is static by doing the following:
- Examined all libraries in lib/lib with `for f in *; do { ldd $f; } done | grep stdc;` - no output was produced.
- Checked for libstdc++ symbols in librdag using nm -a and inspecting the symbols by eye.
- Run on the HPC dev box that alerted us to the issue - native library linkage appears to be working correctly using the system-installed Java binary with no modules loaded. (For the PR record, this is a CentOS 6 box running OpenJDK 1.7.0_55 installed from the yum repository).

This is enough to convince me that this change is having the intended effect, so I am happy to merge this.

PR #124, opened 2014-07-25
[MAT-450] Add support for Windows 32-bit

gmarkall:
This PR along with two others in OG-Lapack and OG-Izy adds support for 32-bit Windows to OG-Maths.

No changes to the native code were made - however, there were some changes to the Java code to support loading more variants of the libraries. Other changes were to the build system and properties files. Please see commit messages for details of individual changes.

This PR also include a fix to the Javadocs that caused an error with doclint on JDK8 - I noticed this because I am using JDK8 on my Windows VM.

Coverage: unchanged


stuartarchibald:
Thanks for the change set. Nice to see it worked on w32 with very little effort.
I'm adding:
- [MAT-456](http://jira.opengamma.com/browse/MAT-456) to make {win,w32,w64},{osx,mac},{lnx,lin} consistent!
- [MAT-457](http://jira.opengamma.com/browse/MAT-457) to fix whatever it is in the native writer that is causing it to fail on w32.

Happy to merge.

PR #126, opened 2014-08-07
[Mat-426] Wiring in OG-Izy and using it in working implementations.

stuartarchibald:
This PR makes a start on all the plumbing, testing and wiring required for the OG-Izy style vector functions. Unfortunately, this does mean it includes the addition of quite a bit of code that is added all at once due to the overhead caused by doing such a task in a fragmented manner. Similarly with the code generators, they are work in progress, a number of things had to change simultaneously and the code below contains the most basic of use cases and some sample implementations. I'm of the view that most will need refactoring, however I wanted to get this code in as it's getting quite large and it is unknown how much more effort is need to work out exactly how to make the various combinations of tooling do the thing I want!

This PR:
- Adds in the template headers for `OG-Izy` functions and most of the partial template specialisation required to wire them up in a type invariant manner.
- Adds in the generated code for the "safe" (by the Sethi-Ullman discussion definition) nodes, i.e. ones that don't change mathematical domain when run.
- Adds in autogenerated tests for the above, these are provided by m-code and are committed to the source as having a dependency on an interpreter of m-code at compile time is undesirable. At the minute only the C++ side testing is generated. The java side will be generated as part of the next refactor.
- Adds in jinja2 templating capability to the node generator. This was to deal with needing flow control in templates as some OG-Izy functions are order invariant e.g. `addx()` has no `xadd()` counterpart as it would be pointless. However e.g. `subx()` has a `xsub()` counterpart as clearly necessary. As a result of this the function prototypes are different and so some flow control is needed at generation time to perform introspection on the functions provided and then order the arguments at the call site correctly.
- Adds in the common infix operator nodes and adds hand written tests for them from DOGMA and java.
- Adds in the capability of specifying tolerances in the `CheckNode()` classes.

As noted in offline discussion, the tolerances in the autogenerated code for the OG-Izy bound functions are somewhat high. This is due to 32bit windows defaulting to mscrt maths library functions which differ in result to those provided by 64bit/POSIX. A workaround is slated and will be added in due course.

Coverage:
Java: 82% (up 5%)
build/src/librdag 40.8% (up 23.7%)
src/jshim 53.1% (no change)
src/librdag 85.7% (down 8.5%, this is partly due to the really simple dispatch stubs not being covered for the autogenerated stuff, I'll look into adding this in as part of the template based generation. Also it's partly due to the addition of `izy.cc` which is entirely added and implemented, but not yet entirely used as noted above.)
src/librdag/runners 94.8% (no change).


gmarkall:
This looks good. Some general remarks:
- Thanks for adding the tests that increase the coverage of the `build/src/librdag` folder! It is something we've needed to add coverage to for a while, and this is a big step in the right direction.
- Thanks for adding the Jinja2 support - as well as making it easier to write the specific Izy implementations, I imagine it can also be ued to simplify some of the existing code generation, which will be desirable. I've created [MAT-461](http://jira.opengamma.com/browse/MAT-461) for this to be picked up when one of us has some free time to work on it.
- Replacing the direct calls to maths library functions makes sense - as discussed in the offline discussion, it does open up the opportunity for making the computations more consistent across platforms (particularly Windows 32-bit vs. everything else) by only changing the maths library backing of OG-Izy. 
- Having the ability to specify the tolerances in the `CheckNode()` calls is also very nice, even if the reason that it needed to be added in this case was due to the Windows 32-bit discrepancies - I imagine it will be useful in the future for good reason when we have tests of more and more complicated nodes implemented in terms of other nodes.
- The m-code is understood to be in an intermediate form at the moment that will change imminently, so for this PR it is not worth providing exhaustive documentation or doing any refactoring for presentation reasons at this time.
- However, I have checked the execution of the m-code generator manually, and it ran fine on my machine. It seems that some whitespace was different (perhaps due to my version of octave) but I don't think this is an issue.

In summary - this looks good, and I'm happy to merge.

PR #128, opened 2014-08-22
[Mat-462] Add checking capability for inputs to LAPACK 

stuartarchibald:
This PR is purely to add in some code/idea stubs because I don't want to go any further with it before we decide it's sensible given the number of call sites and bits of logic that will need adjusting due to the size of the LAPACK namespace.

The intent is:
- To add a template parameter to LAPACK namespace routines that are susceptible to getting stuck (xgesvd) or just producing nonsense on the presence of non-finite data. The template parameter is added to such routines to specify what is to be checked for in the input data prior to the call the LAPACK. At the minute `nothing` and `isfinite` are implemented. The decision to add the option as a template parameter opposed to the method prototype is so that the prototypes remain LAPACK like, also to avoid clumsy branching on testing types in the LAPACK call staging implementation. This also means that in routines where many LAPACK calls are made, necessary checks can be made prior to LAPACK calls which can then proceed without further checking.
- The LAPACK calls are all template functions so the necessary partial template specialisation is performed via buffering the call site through a specialisation struct.
- Adds two new exceptions that extend `rdag_error`, these are `rdag_recoverable_error` (error present but is recoverable) and `rdag_unrecoverable_error` (error present and cannot recover). These are added in part to circumvent the need to check many LAPACK routines' `info` variable status, and also in part to just make the exception hierarchy more sensible. Should this seem sufficient I'll mirror this on the Java side.
- Fixes up a load of runner code to use smart pointer based memory management, I did this whilst adjusting the LAPACK call sites.

Coverage:
Java: 82% (no change)
build/src/librdag 40.8% (no change)
src/jshim 53.1% (no change)
src/librdag 85.9% (up 0.2%)
src/librdag/runners 95.8% (up 1%).

### DIFF
--- a/src/librdag/runners/norm2runner.cc
+++ b/src/librdag/runners/norm2runner.cc
@@ -75,19 +75,12 @@ norm2_dense_runner(RegContainer& reg, shared_ptr<const OGMatrix<T>> arg)
     {
       lapack::xgesvd<T,lapack::OnInputCheck::isfinite>(lapack::N, lapack::N, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
     }
-    catch (exception& e)
+    catch (rdag_recoverable_error& e)
     {
-      try
-      {
-        throw;
-      }
-      catch (rdag_unrecoverable_error& e)
-      {
-        throw;
-      }
       // error is recoverable so just complain
       cerr << e.what() << std::endl;
     }
+    // Else, exception propagates, stack unwinds
 
     ret = OGRealScalar::create(std::real(S[0]));
   }
--- a/src/librdag/runners/norm2runner.cc
+++ b/src/librdag/runners/norm2runner.cc
@@ -68,7 +68,7 @@ norm2_dense_runner(RegContainer& reg, shared_ptr<const OGMatrix<T>> arg)
     std::memcpy(A, arg->getData(), sizeof(T)*m*n);
 
     // call lapack
-    lapack::xgesvd(lapack::N, lapack::N, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
+    lapack::xgesvd<T,lapack::OnInputCheck::isfinite>(lapack::N, lapack::N, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
 
     ret = OGRealScalar::create(std::real(S[0]));
     delete[] A;
--- a/src/librdag/runners/norm2runner.cc
+++ b/src/librdag/runners/norm2runner.cc
@@ -61,18 +61,35 @@ norm2_dense_runner(RegContainer& reg, shared_ptr<const OGMatrix<T>> arg)
 
     T * U = nullptr; //IGNORED
     T * VT = nullptr; //IGNORED
-    real8 * S = new real8[minmn];
+
+    std::unique_ptr<real8[]> Sptr(new real8[minmn]);
+    real8 * S = Sptr.get();
 
     // copy A else it's destroyed
-    T * A = new T[m*n];
+    std::unique_ptr<T[]> Aptr(new T[m*n]);
+    T * A = Aptr.get();
     std::memcpy(A, arg->getData(), sizeof(T)*m*n);
 
     // call lapack
-    lapack::xgesvd<T,lapack::OnInputCheck::isfinite>(lapack::N, lapack::N, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
+    try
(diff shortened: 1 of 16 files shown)


src/librdag/runners/norm2runner.cc
```diff
@@ -61,18 +61,35 @@ norm2_dense_runner(RegContainer& reg, shared_ptr<const OGMatrix<T>> arg)
...
+    T * A = Aptr.get();
     std::memcpy(A, arg->getData(), sizeof(T)*m*n);
 
     // call lapack
-    lapack::xgesvd(lapack::N, lapack::N, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
+    try
+    {
+      lapack::xgesvd<T,lapack::OnInputCheck::isfinite>(lapack::N, lapack::N, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
+    }
+    catch (exception& e)
```
gmarkall: i think this can be reorganised into:

```
try { ... }
catch (rdag_recoverable_error& e) { ... }
catch (rdag_unrecoverable_error& e) { ... }
```

(At least I think that is the intent of this code - if it is, then I think this change would make it clearer).

src/librdag/runners/svdrunner.cc
```diff
@@ -46,18 +46,20 @@ template<typename T> void svd_dense_runner(RegContainer& reg, shared_ptr<const O
   // call lapack
   try
   {
-    lapack::xgesvd(lapack::A, lapack::A, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
+    lapack::xgesvd<T, lapack::OnInputCheck::isfinite>(lapack::A, lapack::A, &m, &n, A, &lda, S, U, &ldu, VT, &ldvt, &info);
   }
   catch (exception& e)
   {
-    if(info < 0) // illegal arg
+    try
```
gmarkall: I think this try/catch can also be reorganised as above.

src/librdag/test/check_iss.cc
```diff
@@ -141,3 +141,35 @@ TEST_F(issTest, isMatrix)
...
 }
+
+
+TEST(issTest_isfinite, real8)
+{
+  std::unique_ptr<real8> rd (new real8[4]);
+  real8 * r = rd.get();
+  r[0]=1;
+  r[1]=2;
+  r[2]=std::numeric_limits<real8>::signaling_NaN();
```
gmarkall, line 152: Would it be worth checking with quiet NaNs as well? I seem to recall that things like `1/0` will produce quiet NaNs, and we'll still want to be able to detect these.

stuartarchibald, line 152: Fair point and I'll add tests for completeness/API demonstration, but I believe that `std::isfinite()` is required to catch C's `FP_NAN`, which encompasses both quiet and signalling variants. I think it's designed this way such that the test is very quick as it can be done with just a bit mask.

src/librdag/runners/norm2runner.cc
stuartarchibald: Thanks for spotting this, I agree with the notion, it does need attention. From offline chat, we agree on the following statement organisation:

```
try
{
 lapack::someExampleCall();
}
catch (rdag_recoverable_error& e)
{
  // error is recoverable so just complain
  cerr << e.what() << std::endl;
}
// Else, exception propagates, stack unwinds
```

I will patch this into the offending sites.

gmarkall:
I think this looks generally good:
- Thanks for adding the additional documentation of template parameters in many places
- The changes from `new[]` to `unique_ptr` in the code of several runners is also a step forward
- I think `nothing` and `isfinite` are good identifiers, since they are descriptive and leave room for other checks to be added without being inconsistent (contrasted with just `yes` or `no`, etc.).
- The requirement to use `PTSBuffer` does suggest that we could reduce a level of indirection through function calls if we made the Lapack interface a class (I think this was part of our discussion before, and maybe we need to discuss it a bit more).
- The split of RDAG errors into recoverable and unrecoverable makes sense.
- Regarding the comment about signaling/quiet NaNs in testing - I stand corrected on this, this will make no difference.

stuartarchibald:
Many thanks for reviewing. I found another couple of places with the weird `try...catch(){}` setup you pointed out (`lapack::xgetrf()` [LU decomp] call side related), so patched them too. I've also added demo tests for `qNaN`s. All fixups are in 69af1e3.

gmarkall:
Thanks for the fixups - they all do the job. Merging.

PR #129, opened 2014-09-10
Mat 468: patch up sources following build system upgrade

stuartarchibald:
This minor patch set is the result of weeks of work upgrading the entire build system.
We now have dedicated machines for building the 64 and 32 bit windows variants.
GCC is globally upgraded to version 4.9.1.
CMake is globally upgraded to 3.0+.

The native library loader and assembler system is patched with regards to the new GCC. Whilst doing this I unified the OS labelling scheme across all sources as:

```
"w64" = 64bit Windows
"w32" = 32bit Windows
"lnx" = Linux
"osx" = OSX
```

It also transpired in the upgrade of GCC that there were a few instances of signed/unsigned int comparison, I have duly patched them. It also appears the Google's gtest code does not compile with `-Wsign-compare` which is applied via `-Wall`. I tried upgrading gtest to version 1.7, the same issue is present. I've therefore added in an override to `-Wall` of `-Wno-sign-compare` and noted it as something we need to look at [MAT-469](https://opengamma.atlassian.net/browse/MAT-469).

Code coverage:
No material change. The minor edits are all in system level places.


CMakeLists.txt
```diff
@@ -98,7 +92,7 @@ set(PATH_AND_JAR_JAVADOC_NAME "${JARDIR}/${JAR_JAVADOC_NAME}" )
 
 # Use C99 and C++11, other flags for preventing common errors. Also add thread libs.
 set(CMAKE_C_FLAGS "-std=c99 -Wall -Werror -Wextra -pedantic ${CMAKE_THREAD_LIBS_INIT} ${CMAKE_C_FLAGS}")
-set(CMAKE_CXX_FLAGS "-std=c++11 -Wall -Werror -Wextra -pedantic ${CMAKE_THREAD_LIBS_INIT} ${CMAKE_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS "-std=c++11 -Wall -Wno-sign-compare -Werror -Wextra -pedantic ${CMAKE_THREAD_LIBS_INIT} ${CMAKE_CXX_FLAGS}")
```
gmarkall, line 95: Given that this is to get gtest to compile, can it be pushed into the gtest CMake stuff so it doesn't get applied to the rest of our codebase? I think a suitable place would probably be in its `config_compiler_and_linker` macro in `cmake\internal_utils.cmake`.

gmarkall:
Thanks for all the effort that went into this - the diff isn't big but I know the whole effort has been huge and frustrating. 

I think this looks good to merge, but will wait on the delpoy builder finishing testing to make sure that all is sound with the modified platform codes. Your comments on moving the flag to gtest (in private chat) are understood - this is a temporary workaround  and this is the straightforward way around the issue, so I am happy with this.

Will merge on the buildbot deploy queue pass.

PR #132, opened 2014-09-25
MAT-475 Fix OSX libstdc++ linkage issues

stuartarchibald:
Testing [MAT-474](https://opengamma.atlassian.net/browse/MAT-474) (assemble license docs into jar) highlighted that the OSX build is suffering from a problem where libraries linked with GCC's `-static-libc++` are still being dynamic linked to `libstdc++`. This ended up being a massive trudge through GCC and OSX system internals, the eventual fix is a return to dynamic linking on OSX only.

This changeset:
- Updates the GCC location on the OSX syslib builder as GCC 4.9.1 is now the version unified across the entire build system.
- Fixes up some instances of catching by value opposed to reference.
- Adds a missing `toString()` to `ComplexArrayContainer` that was needed during debugging.
- Switches OSX to use dynamic linking for `libstdc++` and to link using the 2 layer namespace.

A more thorough explanation for the last item is given in the commit message of e2e97ba. Basically, GCC bug, a few possible visibility bugs, a leaky system library and a raft of unknowns makes dynamic linking for `libstdc++` the sensible option on OSX at the present time.

Coverage:
No change worth reporting.


gmarkall:
Thanks for patching up those other minor various issues along with this pull request.

I agree with everything you've written - whilst it would be ideal for every system to be built in the same way, it is not really practical until we have a better understanding of linking on OS X (and even with a greater understanding, we may only come to understand why it's better this way).

This looks good to merge.

PR #133, opened 2014-09-25
MAT-474 add licenses into jar build

stuartarchibald:
This PR, to be read in conjunction with https://github.com/OpenGamma/OG-Izy/pull/12 and https://github.com/OpenGamma/OG-Lapack/pull/15:
- Adds the approved license directory structure into all jars and then from the linux jar into the combined jar.
- Forces the build dependencies such that items created by the `NativeLibraryBuilder` cannot be built without assembling the licenses correctly.
- Adds code to import licenses from OG-Izy and OG-Lapack.
- Fixes some minor javadoc issues that were causing compilation failure locally.

Coverage:
No change.


gmarkall:
This looks good to me - happy to merge.

PR #136, opened 2014-09-30
Mat 476 adding deploy target

stuartarchibald:
This PR:
- Adds a custom maven deployment test target. This is now called by the build system as the final step to check that the provided `pom.xml` is at least parseable!
- Specifies a version of slf4j required for the OG-Maths jar.

Coverage:
No change.


gmarkall:
This looks good. I've tried it with the correct pom as-is, and also introducing the mistake on the first line of writing `<project>` as opposed to `<project`, which causes this new target to fail as expected.

PR #137, opened 2014-10-01
[MAT-478] Quieten mldivide impl.

stuartarchibald:
This makes the mldivide implementation quieter. Until there's a proper logger and log drain, all statements (apart from a couple of important diagnostics) are wrapped in a guard flag. This flag is controlled by a compile time constant and is set as `false` (no reporting) for all builds except debug.

Coverage:
No change of significance.

### DIFF
--- a/src/librdag/runners/mldividerunner.cc
+++ b/src/librdag/runners/mldividerunner.cc
@@ -35,6 +35,17 @@ namespace detail
 {
 
 /**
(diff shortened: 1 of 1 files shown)


gmarkall:
This looks good - from the build logs, it looks like _dbg builds still print out these logging messages and others don't, which meets the stated goal of this PR.

PR #138, opened 2014-10-02
[MAT-479] Add requirement for finite system to mldivide.

stuartarchibald:
This adds a scan of the system matrix to the mldivide function.
This scan will cause an exception to be raised if the matrix
contains non-finite data. The reason for this behavior is to
prevent LAPACK solvers receiving data containing inf/NaN as it
makes them do rightly bizarre things.

Coverage:
Java: 82% (no change)
build/src/librdag 40.8% (no change)
src/jshim 53.1% (no change)
src/librdag 85.9%  (no change)
src/librdag/runners 96.1% (up 0.3%).


gmarkall:
This looks good to me.

